PR #18, opened 2025-05-27
Fixed HoverCardContext's hovercard clips over the edge of the page

Affan-mulla:
Now on hovering the card will not leave the screen

### DIFF
--- a/src/generic/HoverCardContext.tsx
+++ b/src/generic/HoverCardContext.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useContext, useState } from 'react';
+import React, { createContext, useContext, useLayoutEffect, useRef, useState } from 'react';
 import './styles.css';
 
 type HoverCardData = {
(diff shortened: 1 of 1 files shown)


src/generic/HoverCardContext.tsx
```diff
@@ -32,11 +32,38 @@ export const HoverCardProvider: React.FC<{ children: React.ReactNode }> = ({ chi
...
+      setHoverCard((prev) => ({ ...prev, x: newX - 10, y: newY - 10 }));
+    }
+  }, [hoverCard.x, hoverCard.y, hoverCard.visible]);
+
   return (
     <HoverCardContext.Provider value={{ hoverCard, showHoverCard, hideHoverCard }}>
       {children}
       {hoverCard.visible && (
         <div
+         ref={cardRef}
```
conradarcturus, line 66: The tab spacing looks off. On my computer the linter fixes it. Not sure if you had it installed when you made the change. There's a few other things lint rules like not having 2 newlines in a row. However that sounds like splitting hairs and I appreciate the change :)

Nonetheless everything works! So I'm happy to just land this as-is if its not easy to run the linter.

conradarcturus (approved):
image

Issue #6, opened 2025-05-19
Add IANA Variant object type


conradarcturus:
Locale codes usually look like "es_MX" Spanish in Mexico. Sometimes there is script information -- eg. zh_Hans_CN, the "Hans" means "in the Simplified Han writing system". Even rarer, there are variant tags, like ca_ES_VALENCIA or rm_CH_RUMGR. Right now we can intake that information for locale codes, but the data does not have much context.

The IANA variant subtags (see https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry 

For a locale like `rm_CH_RUMGR` the `rm` part is a language and it has an object `LanguageData` . The `CH` part (Switzerland) is a territory and has an object `TerritoryData`. The goal of this task is to make a new object type that has `VariantTagData` so `RUMGR` is saved with more data.

In order to do that you need to:
1. Update `ObjectType` enum in `DataTypes.tsx` to add a new value: VariantTag
2. Add a routine to load it in `CoreData.tsx` -- probably good to write that function in a new file in that folder `IANAData.tsx`. You have an open PR where you started working with that data -- I recommend finishing that before starting this issue.
3. Now, more than just modifying locales, we will be adding actual Variant data. So make a datatype `VariantTagData` in `DataTypes.tsx` -- the fields will be explained below.
4. We will need to save the `variants` to the `CoreData` that `useCoreData` generates. Look at how we add locales, writing systems, and other object types to see how that can be done.
5. You'll want to associate the variant information to locale objects as well. So just like how there is a method `connectLocales` you will want to make a method `connectVariantTags`. We'll want to connect VariantTag objects with languages and locales they are relevant to. You'll want to add `variantTags: VariantTagData[]` to the `LanguageData` type and `variantTag?: VariantTagData` to the `LocaleData` type.
6. When rendering a locale in `LocaleDetails.tsx` you can now show the VariantTag object data. In order to do that you should add a new `VariantTagCard.tsx` component similar to the `LocaleCard.tsx` -- that's probably a good time to stop the PR. We can make subtasks for making the separate view components (CardList, Table...).

The `VariantTagData` object should look like this:

```tsx
export interface VariantTagData extends ObjectBase {
  type: ObjectType.VariantTag;
  ID: VariantIANATag;
  codeDisplay: VariantIANATag;
  nameDisplay: string;

  description: string;
  associatedLanguageCodes: LanguageCode[];

  // References to other objects, filled in after loading the TSV
[...]

PR #29, opened 2025-05-29
3 add locales for languages with iana variants

wizardsimms:
close


src/dataloading/CoreData.tsx
```diff
@@ -54,9 +55,6 @@ export const EMPTY_LANGUAGES_BY_SCHEMA: LanguagesBySchema = {
   CLDR: {},
 };
 
-/**
```
conradarcturus, line 57: Is there a reason you removed this comment?

src/dataloading/CoreData.tsx
```diff
@@ -103,13 +104,14 @@ export function useCoreData(): {
     addISOLanguageFamilyData(languagesBySchema, langFamilies || [], isoLangsToFamilies || {});
     addISOMacrolanguageData(languagesBySchema.ISO, macroLangs || []);
     addGlottologLanguages(languagesBySchema, glottologImport || [], manualGlottocodeToISO || {});
-    addCLDRLanguageSchema(languagesBySchema, cldrAliases || []);
+    addCLDRLanguageSchema(languagesBySchema, cldrAliases || {});
```
conradarcturus, line 107: is there a reason you replaced the array with an object here?

src/dataloading/IANAData.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  return variants;
+}
+
+export async function addIANAVariantLocales(
+  languagesBySchema: LanguageSchemas,
+  locales: Record<string, LocaleStub>
+): Promise<void> {
+  const variants = await loadIANAVariants();
```
conradarcturus, line 44: instead of making this async, it looks like you are already passing in an array of type `IANAVariantData[]`

src/dataloading/IANAData.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  for (const variant of variants) {
+    for (const prefix of variant.prefixes) {
+      const cldrLang = languagesBySchema.CLDR[prefix];
+      if (!cldrLang) continue;
+
+      const iso639_3 = cldrLang.ID;
+      const localeCode = `${prefix}_${variant.tag}`;
+
+      locales[localeCode] = {
```
conradarcturus, line 54: Check first if locales[localeCode] does not exist yet.

src/dataloading/IANAData.tsx
```diff
@@ -0,0 +1,65 @@
...
+        ID: localeCode,
+        Language: iso639_3,
+        Variant: variant.tag,
+        Name: variant.name,
+        Scope: ScopeLevel.Parts,
+        Canonical: false,
```
conradarcturus, line 60: This is looking very promising. ID and scope look good. All field (except ID) should be camelCase.

* Language should be a reference to an object, `language: cldrLang` but you also want the string `languageCode: cldrLang.ID`
* Variant should be `variantTag: variant.tag`. We'll add `variant` object in a different issue/PR -- for now you are just adding new locales
* Name needs to combine the language name & variant name, eg `nameDisplay: cldrLang.nameDisplay + " (" + variant.name + ")"`
* `codeDisplay` needs to be defined -- that is the localeCode as well.
* `type` is ObjectType.Locale
* `territoryCode`, `populationSource` and `populationEstimate` are all empty but currently required. Rather than worrying about the full refactor you can just use empty strings or 0 for now. We can change those types later.

For full reference, the LocaleData type is defined here: 

https://github.com/Translation-Commons/lang-nav/blob/9db55c2e28801191c22ff8a12987a921f4e0007a/src/types/DataTypes.tsx#L122-L145

conradarcturus (changes requested):
Great start! Thanks for beginning this.

src/dataloading/IANAData.tsx
conradarcturus: Something seems out of sync, you pass in ianaVariants with CoreData.tsx but here you aren't. Did you change this file but forget to stage the change to the new commit?

src/dataloading/IANAData.tsx
conradarcturus: Still anticipating this change and the one below

conradarcturus (changes requested):
Looks like you still need to update IANAData.tsx. Perhaps you did but you forgot to add the change to the commit. What developer environment do you use? Perhaps I can give you some tips.

I use VSCode and I manage version control usually with the Interactive Git Log interface. I fall back to terminal if I need to do something the git log cannot do.

src/data/IANAData.tsx
```diff
@@ -0,0 +1,65 @@
+import { ScopeLevel, LocaleStub, LanguageSchemas } from './Types';
```
conradarcturus, line 1: You are importing things from a non-existent file.

ScopeLevel is imported from 
```
import { ScopeLevel } from '../types/ScopeLevel';
```

Others from other files. If you have set up your IDE correctly, then your IDE can automatically figure out the correct imports.

src/data/IANAData.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  return variants;
+}
+
+export async function addIANAVariantLocales(
+  languagesBySchema: LanguageSchemas,
+  locales: Record<string, LocaleStub>
+): Promise<void> {
+  const variants = await loadIANAVariants();
```
conradarcturus, line 44: Don't use the awaitable in this function. Remove it and pass in variants.

src/data/CoreData.tsx
```diff
@@ -59,6 +60,7 @@ export const EMPTY_LANGUAGES_BY_SCHEMA: LanguagesBySchema = {
 /**
  * Get core data needed to show the tables -- things like language codes, relationships with other languages.
  */
+
```
conradarcturus, line 63: Remove this unrelated change
```suggestion
```

src/data/CoreData.tsx
```diff
@@ -116,6 +121,7 @@ export function useCoreData(): {
     connectWritingSystems(languagesBySchema.Inclusive, territories, writingSystems);
     connectLocales(languagesBySchema.Inclusive, territories, writingSystems, locales);
     createRegionalLocales(territories, locales); // create them after connecting them
+    addIANAVariantLocales(languagesBySchema, locales, ianaVariants);
```
conradarcturus, line 124: lgtm, you can simplify it a little bit by passing in the CLDR language list.
```suggestion
    addIANAVariantLocales(languagesBySchema.CLDR, locales, ianaVariants);
```

conradarcturus (approved):
Thanks for dedicating the time to walk through the change together. It looks good! There are some follow-ups we will want to do to add more variant data -- but we will do that in with the next issue: https://github.com/Translation-Commons/lang-nav/issues/6

PR #62, opened 2025-06-11
Add LocaleParser.ts

wizardsimms:
Issue #6


conradarcturus:
Thanks for posting so many PRs :D I don't quite understand this one -- what's the point of this one? Please add a description to this PR, especially before tagging reviewers (since they will all get emailed).

Note I added a github action that now automatically rebuilds the code when you post a PR -- it looks like this change is also causing some build failures. I recommend investigating that.

PR #71, opened 2025-06-12
Update PopulationData.tsx

wizardsimms:


conradarcturus:
Unfortunately the issue cannot be resolved right now. The problem is that in order to convert older population numbers we do not have the data in the repo. For example, if it says 15,581,334 English speakers in Australia to modern numbers we can only convert it to 2025 numbers if we know either 1) the year that estimate is from (in this case 2006) or 2) the percent of the population (in this case 78.47%). I'm changing the data to be loaded from census information.

By the way, I strongly recommend commenting on issues and messaging me when things don't add up. The changes you posted contain numerous errors and invent data that does not exist. I recommend reading over the codebase more as well so you don't invent values. I'm a bit afraid for making this change you are relying on GenAI tools and they are hallucinating a lot.

PR #81, opened 2025-06-17
Update PageParamsContext.tsx

YashwanthKumar49:
This update introduces a new URL parameter management function (replaceAllParams) and enhances the robustness of existing parameter parsing within the PageParamsContext. These changes are foundational, enabling more precise control over the application's URL state.

### DIFF
--- a/src/controls/PageParamsContext.tsx
+++ b/src/controls/PageParamsContext.tsx
@@ -15,6 +15,7 @@ import { ScopeLevel } from '../types/ScopeLevel';
 
 type PageParamsContextState = PageParams & {
   updatePageParams: (newParams: PageParamsOptional) => void;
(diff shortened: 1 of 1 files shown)


conradarcturus (changes requested):
This is a patch but there's a bigger problem to solve. I think the `ObjectTypeSelector` should use `<a href={getNewURL(...}}><button...></a>` instead. That way people can open new windows by right-clicking on the buttons (or also use the buttons to copy URLs).

We will need to create a class with a name like `HyperlinkChoiceOptions` that's a copy of `SingleChoiceOptions` with the main difference being that it uses `a href` rather than `button onClick`.

conradarcturus:
Closing since its redundant to https://github.com/Translation-Commons/lang-nav/pull/93

PR #93, opened 2025-06-29
Update ObjectType Selector

YashwanthKumar49:
When a user switches between object types (e.g., from Census to Locale), we want to:

Clear any previous search filters (searchString: undefined)

Reset pagination to the first page (page: 1)


conradarcturus (approved):
image

PR #94, opened 2025-06-30
Added column filtering for tables #91

### LINKED ISSUE #91: UX: Add Column filtering
We have a really useful class `ObjectTable.tsx` that can show various tables of information.

Over time, we expect to add more columns of data. As the table gets wider and wider it will be harder to use. Fortunately, we can mitigate that by adding options for people to turn off columns.

We'll add toggle options at the bottom of the "Table" pages just like we added to the Hierarchy page in https://github.com/Translation-Commons/lang-nav/pull/55 

Image

# Suggestions
Here's some recommendations to go about this:
1. In `ObjectTable.tsx` there is an interface `TableColumn` that specifies that the key is an optional string -- we will want to use this key for the visibility filter. But we need it to be required, remove the `?` to make it required. 
    1.  We can set the `label` param as optional and when `label` is used but the label is null, use the key, like: `label ?? key` 
    2. That will cause a lot of errors since currently none of the columns specify the column key -- so go through all of the usages of TableColumn.  If the label is a plain string, just change that param to a key. If the label is markup (for example, has a hoverable), make a separate key.
2. Now that we have keys, we want to create a state variable to keep track of check columns are toggled. You'll do this in `ObjectTable.tsx`
    1. Use `useState` to create an object mapping strings to booleans -- useState will provide 2 variables, `visibleColumns` and `setVisibleColumns`
    2. You should set all columns to be visible initially, you should be able to initialize that with `Object.fromEntries(columns.map(col => [col.key, true]))`
    3. Make a function using React's `useCallback` feature `toggleColumn` that will take a string input. This will modify visibleColumns -- but not by mutating visibleColumns directly, rather by using `setVisibleColumns`. You'll want a function like `setVisibleColumns(prev => ({...prev, columnKey: !prev[ColumnKey])`. By using the prev argument we are able to avoid race conditions with setting visibleColumns.
3. Put a group of checkboxes at the bottom of the table (or any place you recommend if you have a good idea). For each column you will have a map
4. Then when you are rendering columns (around `column.render(object)`) you will make a conditional statement checking if the column is visible. If its not, then return null to skip the table cell for that column.

## Example code

Making multiple elements for each column that was passed in
https://github.com/Translation-Commons/lang-nav/blob/0691c6d0bbe28724ba3af1b1f74b2cd340b1ac8d/src/views/common/table/ObjectTable.tsx#L62-L70

Checkbox example
https://github.com/Translation-Commons/lang-nav/blob/0691c6d0bbe28724ba3af1b1f74b2cd340b1ac8d/src/views/common/TreeList/TreeListOptions.tsx#L81-L88

Pranav-Vishwanatham:
This pull request implements the column filtering feature for the `ObjectTable` component, resolving Issue [#91](https://github.com/Translation-Commons/lang-nav/issues/91).

**Changes made:**
- Updated the `TableColumn` interface in `ObjectTable.tsx` to require a unique `key` property.
- Added state management within `ObjectTable.tsx` to track and toggle column visibility.
- Added a checkbox UI to allow users to control which columns are shown.
- Updated all existing tables (`Languages`, `Censuses`, etc.) to provide the required `key` for each column definition.

Fixes #91

### DIFF
--- a/src/views/common/table/ObjectTable.tsx
+++ b/src/views/common/table/ObjectTable.tsx
@@ -72,8 +72,11 @@ function ObjectTable<T extends ObjectData>({ objects, columns }: Props<T>) {
         nOverall={objects.length}
         objectType={objects[0]?.type}
       />
-      <div className="column-toggler" style={{ margin: '1rem 0', display: 'flex', flexWrap: 'wrap', gap: '1rem' }}>
-        <strong>Show/Hide Columns:</strong>
+      <details className="column-toggler" style={{ margin: '1rem 0', gap: '1rem' }}>
+        <summary className="collapsible-summary" style={{ cursor: 'pointer' }}>
+          {currentlyVisibleColumns.length}/{columns.length} columns visible, click here to toggle.
+        </summary>
+        <div className="column-toggler" style={{ margin: '1rem 0', display: 'flex', flexWrap: 'wrap', gap: '1rem' }}>
         {columns.map(column => (
           <label key={column.key} style={{ cursor: 'pointer' }}>
             <input
@@ -85,7 +88,8 @@ function ObjectTable<T extends ObjectData>({ objects, columns }: Props<T>) {
             {column.label ?? column.key}
           </label>
         ))}
-      </div>
+        </div>
+      </details>
 
       <table className="ObjectTable">
         <thead>
--- a/src/views/common/table/ObjectTable.tsx
+++ b/src/views/common/table/ObjectTable.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction, useMemo, useState } from 'react';
+import React, { Dispatch, SetStateAction, useMemo, useState, useCallback } from 'react';
 
 import { getScopeFilter, getSliceFunction, getSubstringFilter } from '../../../controls/filter';
 import { usePageParams } from '../../../controls/PageParamsContext';
@@ -11,9 +11,9 @@ import VisibleItemsMeter from '../../VisibleItemsMeter';
 import './tableStyles.css';
 
 export interface TableColumn<T> {
-  label: React.ReactNode;
+  label?: React.ReactNode;
   render: (object: T) => React.ReactNode;
-  key?: string;
+  key: string;
   isNumeric?: boolean;
   sortParam?: SortBy;
 }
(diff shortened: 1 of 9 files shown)


src/views/language/LanguageTable.tsx
```diff
@@ -23,16 +23,19 @@ const LanguageTable: React.FC = () => {
...
         },
         {
+          key: 'Population', 
           label: 'Population',
           render: (lang) => lang.populationCited,
           isNumeric: true,
           sortParam: SortBy.Population,
         },
         {
+          key: 'InternetTechnologies',
```
conradarcturus, line 38: you can leave the space in for the key, it doesn't need to be camelcase, just a unique string.

Pranav-Vishwanatham:
Hi Conrad, 

Thank you for suggesting the changes, I have fixed those changes and below are the screenshots of how the new collapsible hoverable menu looks:

Screenshot 2025-07-01 at 1 03 33 PM

image

image

Learnt a lot of things from this task, felt like a really good first issue to work on, thanks for the suggestions.  Looking forward for further more tasks like these. 🙂

src/views/common/table/ObjectTable.tsx
```diff
@@ -56,12 +72,31 @@ function ObjectTable<T extends ObjectData>({ objects, columns }: Props<T>) {
         nOverall={objects.length}
         objectType={objects[0]?.type}
       />
+      <details className="column-toggler" style={{ margin: '1rem 0', gap: '1rem' }}>
+        <summary className="collapsible-summary" style={{ cursor: 'pointer' }}>
```
conradarcturus, line 76: [nit, you don't need to change it right now] you are using classes as well as inline styles. I sometimes use inline styles out of laziness / its never going to be reused. But for this you might as well define the style components all in the CSS.

src/views/common/table/ObjectTable.tsx
```diff
@@ -56,12 +72,31 @@ function ObjectTable<T extends ObjectData>({ objects, columns }: Props<T>) {
         nOverall={objects.length}
         objectType={objects[0]?.type}
       />
+      <details className="column-toggler" style={{ margin: '1rem 0', gap: '1rem' }}>
```
conradarcturus, line 75: [nit, you are welcome to land as-is and ignore and/or change later]: only have a bottom margin. The top margin stacks with your padding for the summary so its a bigger gap than necessary. Also, generally I use `em` instead of `rem`. `rem` is the measured by the default character height for the page, `em` is based on character height for the container its in. So if a container has a smaller character height the `em`s inside change proportionally but the `rem`s would be constant. This website avoids changing font size with nesting so its basically a non-issue here, just for context.
```suggestion
      <details className="column-toggler" style={{ marginBottom: '1rem', gap: '1rem' }}>
```

conradarcturus (approved):
image

Issue #96, opened 2025-07-01
UX: Rename Scope to Granularity


conradarcturus:
The word "scope" works in some cases, but its overused in the tool. I think the word "granularity" fits better. I'm happy to get feedback on this.

PR #99, opened 2025-07-04
Refactor: Scope -> Granularity (#96)

conradarcturus:
The focus of this PR is to improve the comprehension of the "scope" or "granularity" shown. I'm happy to get further feedback. Naming is hard -- so this isn't obvious but I think this is a good step forward. The changes are:

* Rename "scope" to "granularity" in most cases.
* Change names of the levels to Macro, Base, Micro, Special. I'm happy to have further feedback.
* Make a standard "granularity" column for the table view and move it as the first column -- used by Writing Systems and Locales. Languages and territories have finer distinctions so those are left intact. Censuses don't have a granularity distinction today.
* Added a tooltip to explain language scope better

|What|Changes|Before|After|
|--|--|--|--|
|[Locales table for Romansch](https://translation-commons.github.io/lang-nav/?view=Table&objectType=Locale&granularities=Base%2CMacro%2CMicro%2CSpecial&searchString=roh&searchBy=ID&scopes=Individuals%2CGroups%2CParts%2COther)|See the new granularity column to the left|Screenshot 2025-07-03 at 18 59 32|Screenshot 2025-07-03 at 18 58 41
|Scope/Granularity Selector|Name changed|Screenshot 2025-07-03 at 18 59 53|Screenshot 2025-07-03 at 18 55 52
|Tooltip for language granularity|Vastly expanded|Screenshot 2025-07-03 at 19 00 48|Screenshot 2025-07-01 at 09 55 37


conradarcturus:
I decided I was trying to be too central with this idea. In hindsight, I think it's better to break up the idea of LanguageScope & TerritoryScope into separate concepts rather than trying to make a way to merge them.

PR #102, opened 2025-07-09
Replaced unicode characters with Lucide icons (fixes #86)

### LINKED ISSUE #86: UX: Update icons from unicode chars to well-designed web icons
Right now we're using Unicode characters in a few places for icons. However they don't have great visual consistency. We should replace a number of them.

Icons to search for: ✅ ❌ ⚠️ 🔎 A few use unicode code-points rather than direct references `&#x1F50E;` `&#x2197;` ` &#x2139;`

One promising possibility is to use this library https://lucide.dev/ 

Instructions
1. Install Lucide. In your repo's directory, run `npm install lucide-react`
2. In files you want to use them add the import eg. `import { Info, AlertTriangle } from 'lucide-react';` and then use then like below 
   1. `<Info style={{ color: 'var(--color-button-primary)' }} size={18} />` -- note I'm not sure if we can use the color parameter. Our CSS files already define colors that we'd want to reuse.
   2. `<AlertTriangle style={{ color: 'var(--color-text-yellow)' }} size={18} />`
3. Make sure to add their license in the About page `AboutPage.tsx` https://lucide.dev/license

Pranav-Vishwanatham:
This pull request resolves issue #86 by replacing all instances of Unicode characters and HTML entities used as icons with modern, consistent SVG icons from the "lucide-react" library.

### Key Changes

- **Installed `lucide-react`:** Added the new icon library as a project dependency.
- **Replaced Unicode Icons:** Systematically replaced all targeted Unicode characters (e.g., ✅, 🔎, ↗) with their corresponding `lucide-react` components. This affected numerous components, including `CLDRCoverageInfo`, `ViewModal`, `LinkButton`, `SearchBar`, and various tables.
- **Refactored CSS Icon:** Removed a CSS rule that used `content: '↗'` and moved the icon directly into the `LinkButton` component for consistency.
- **Updated About Page:** Added the required license and attribution for the Lucide library to the `AboutPage.tsx`.
- **Corrected `Selector.tsx` Prop Type:** While implementing the new icons, a related bug was fixed. The `selectorLabel` prop in the `Selector.tsx` component was updated from `string` to `React.ReactNode` to resolve a TypeScript error and allow it to accept icon components as labels.

Here are few screenshots:
image
image

Fixes #86

### DIFF
--- a/src/controls/components/TextInput.tsx
+++ b/src/controls/components/TextInput.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useRef, useState } from 'react';
+import { Search, ExternalLink, X } from 'lucide-react';
 
 import HoverableButton from '../../generic/HoverableButton';
 import { View } from '../../types/PageParamTypes';
@@ -118,7 +119,7 @@ const TextInput: React.FC<Props> = ({
           setShowSuggestions(false);
         }}
       >
-        &#x2716;
+        <X size="12px" style={{ color: 'white' }} />
       </button>
     </>
   );
@@ -171,13 +172,13 @@ const SuggestionRow: React.FC<SuggestionRowProps> = ({
     <div className="SuggestionRowWithMultipleInteractions">
       <div>{label}</div>
       <HoverableButton hoverContent={<>Filter by &quot;{searchString}&quot;</>} onClick={setFilter}>
-        &#x1F50E;
+        <Search size="1em" />
       </HoverableButton>
       <HoverableButton
         hoverContent={<>Go to the details page for {searchString}</>}
         onClick={goToDetails}
       >
-        &#x2197;
+        <ExternalLink size="1em" />
       </HoverableButton>
     </div>
   );
--- a/src/controls/selectors/SearchBar.tsx
+++ b/src/controls/selectors/SearchBar.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Search } from 'lucide-react';
 
 import { SearchableField, View } from '../../types/PageParamTypes';
 import Selector from '../components/Selector';
@@ -13,7 +14,7 @@ const SearchBar: React.FC = () => {
   const getSearchSuggestions = useSearchSuggestions();
 
   return (
-    <Selector selectorLabel="🔎">
+    <Selector selectorLabel={<Search size="12px" />}>
       <TextInput
         inputStyle={{ minWidth: '20em' }}
         getSuggestions={getSearchSuggestions}
--- a/src/generic/LinkButton.tsx
+++ b/src/generic/LinkButton.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { ExternalLink } from 'lucide-react';
 
 export default function LinkButton({
   href,
@@ -11,6 +12,7 @@ export default function LinkButton({
     <a href={href}>
       <button className="LinkButton" role="link">
         {children}
+        <ExternalLink size="1em" style={{ marginLeft: '4px' }} />
       </button>
     </a>
   );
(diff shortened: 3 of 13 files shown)


src/controls/selectors/SearchBar.tsx
```diff
@@ -13,7 +14,7 @@ const SearchBar: React.FC = () => {
   const getSearchSuggestions = useSearchSuggestions();
 
   return (
-    <Selector selectorLabel="🔎">
+    <Selector selectorLabel={<Search size="12px" />}>
```
conradarcturus, line 17: Generally we should avoid absolute px measurements for text-like elements. Since 1em=16px then this should be 0.75em.

I'm playing around with other solutions to this though since I think the icon should be bigger (1em) but that's causing unfavorable rendering errors.

src/generic/LinkButton.tsx
```diff
@@ -11,6 +12,7 @@ export default function LinkButton({
     <a href={href}>
       <button className="LinkButton" role="link">
         {children}
+        <ExternalLink size="1em" style={{ marginLeft: '4px' }} />
```
conradarcturus, line 15: lgtm! I don't have a strong rule to use px or em in margins since in cases like this we just want a little gap. Maybe "0.25em" is preferable to "4px" but I think you will find many areas where I was not consistent myself.

conradarcturus (changes requested):
These icons are looking really good! I have some followups but the baseline is definitely well achieved.

Thanks for including the before & after screenshots. When doing it, can you use markdown tables? You can see markdown tables in my PR https://github.com/Translation-Commons/lang-nav/pull/99 -- if you click the button to edit the description you can see the syntax I used.

src/controls/selectors/SearchBar.tsx
Pranav-Vishwanatham:
```suggestion
    <Selector selectorLabel={<Search size="1em" />}>
```

I found that when I set the size to "1em" or "16px", the search icon looks a bit too large and out of proportion with the input field's text, as you can see in the screenshot below.

image

To solve this while still using relative units, I've adjusted the size to "0.75em". I think this size fits the design much better. Let me know what you think!

image

Pranav-Vishwanatham:
Hi @conradarcturus,

Thanks again for all the helpful feedback! I'm still learning, and I really appreciate you taking the time to share these best practices with me.

I've pushed up all the requested changes and updated the PR description with markdown table for the screenshots, as you suggested.

It should be ready for another review when you have a chance.

Thanks!

conradarcturus (approved):
It's good to go. I think we can increase the icon sizes -- but when I tried it on your branch I had issues with margins. So let's go ahead and land this. We can fight the CSS another day.

src/controls/components/TextInput.tsx
```diff
@@ -118,7 +119,7 @@ const TextInput: React.FC<Props> = ({
           setShowSuggestions(false);
         }}
       >
-        &#x2716;
+        <X size="12px" style={{ color: 'white' }} />
```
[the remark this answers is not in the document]
textexin: do any of the icons require localization?
Are they all understood in all cultures?
Will there be any confusion when paired with an RTL language name, etc.?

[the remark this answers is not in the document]
conradarcturus: Thanks for chiming in Tex. In the PRD I listed UI localization as a priority for phase 4. https://docs.google.com/document/d/1z-jhCIBmtUfSxPNII9S3GFdvLA6EmvEQXjWvfZ5dTpU/edit?tab=t.0#heading=h.fjzn37571yva

Keeping it in mind is good, but we need to focus on functional requirements -- I think there are big problems in data and usability that we need to focus on first.

> do any of the icons require localization?

Can you provide references for us to understand the localization challenges of iconography? I bet there are issues I don't foresee -- but I need to understand the issues & compare them to the expected audience of this tool to understand the prioritization.

PR #105, opened 2025-07-11
Style: Add lint and format check

conradarcturus:
Some of the newer devs don't have their environment full set up. This PR adds the ability for github actions to automatically add a commit to PRs to use prettier & es-lint to nicely format the code and standard the code style.


Pranav-Vishwanatham (approved):
Great idea! This is a good addition to the workflow. Automating the code formatting will definitely help keep the style consistent and save time during reviews.  Great idea!

PR #107, opened 2025-07-15
Filters: Separate language & territory scope

conradarcturus:
Take 2: so I'm trying to improve the filter options to make them more understandable. I had an optimistic assumption to add 1 control for multiple purposes: Scope. But that was poorly defined for Censuses and Writing Systems and it can be multiple things for Locales. While I was trying to creating 1 control I realized that there are just 2 different kinds of scope that are used in the tool: Language scope v Territory scope.

|Unitary Scope (old)|Language Scope (new)|Territory Scope (old)|
|--|--|--|
|This would dynamically change depending on the view you were.|Now macrolanguage is separable.|Now you have a lot more granularity. It's not strictly necessary but it shows how the 2 schemas were not compatible.|
|Screenshot 2025-07-15 at 11 02 34|Screenshot 2025-07-15 at 11 02 16|Screenshot 2025-07-15 at 11 02 25

To see how the problem was when we overloaded definitions, it's confusing if we are on a Census details page that we got to from a territory page. Before, the default value was "unit" values (countries, languages (including macrolanguages)) -- so as you see this dropdown doesn't match appropriately -- but now those are separately defined. https://translation-commons.github.io/lang-nav/?view=Details&objectType=Territory&objectID=ca2021.1&scopes=Individuals
|Before|After|
|--|--|
|Screenshot 2025-07-15 at 11 16 17|Screenshot 2025-07-15 at 11 16 13


Pranav-Vishwanatham (commented):
The refactor to separate the filter components looks good to me, and its much clearer and easier to understand than the older version.

I've tested the new filters locally, and I have a question about the functionality. When I'm on the **Languages page** and I use the 'Territory Scope' filter, it doesn't seem to affect the results. My understanding is that this is because the language data itself doesn't contain the territory information needed to filter by.

This made me wonder if it would be a better user experience to only show filters that are relevant to the current page? For example:

On the **Languages page**, we would only show the 'Language Scope' filter.

On the **Territories page**, we would only show the 'Territory Scope' filter.

This would prevent users from trying to use a filter that has no effect. What are your thoughts on this approach? Please let me know if I'm misunderstanding how the data is intended to work.

conradarcturus:
I definitely agree, we don't need to show these filters when they aren't relevant. I rebased this change and did that for the territories scope filter.

However for language scope -- languages appear in every page so I left that filter in everywhere.

PR #110, opened 2025-07-15
Add VariantTag object type

wizardsimms:
This change adds a new object type to example VariantTag information from IANA.

See the new cardlist of variants tags:

And now that they are connected to languages and locales, they now link to variant tag objects.
Screenshot 2025-08-01 at 10 44 12

See how it changes the renders of information in the language & locales details
|Page|Before|After|
|--|--|--|
|[Rumantsch language](https://translation-commons.github.io/lang-nav/data?objectID=roh&view=Details)|Screenshot 2025-08-01 at 10 51 22|Screenshot 2025-08-01 at 10 51 09|
|[Rumasch Grischun Locale](https://translation-commons.github.io/lang-nav/data?objectID=roh_CH_RUMGR&view=Details&objectType=Locale)|Screenshot 2025-08-01 at 10 56 24|Screenshot 2025-08-01 at 10 56 01|


src/views/varianttag/VariantTagCard.tsx
```diff
@@ -0,0 +1,16 @@
+// src/views/varianttag/VariantTagCard.tsx
+
+import { VariantTagData } from '../../types/DataTypes';
+import ObjectTitle from '../common/ObjectTitle';
+import ObjectField from '../common/ObjectField';
+
+export function VariantTagCard({ data }: { data: VariantTagData }) {
```
conradarcturus, line 7: You can largely copy-paste the Census card https://github.com/Translation-Commons/lang-nav/blob/master/src/views/census/CensusCard.tsx -- replacing `Census` with `VariantTag`  

but leave out the "Territory" section and for the languages section you can use the list from the object itself, like from the Writing System Card: https://github.com/Translation-Commons/lang-nav/blob/master/src/views/writingsystem/WritingSystemCard.tsx#L42-L51

src/views/locale/LocaleDetails.tsx
```diff
@@ -78,11 +79,11 @@ const LocaleDefinitionSection: React.FC<{ locale: LocaleData }> = ({ locale }) =
...
       )}
       {variantTag && (
-        <div>
-          <label>Variant:</label>
-          {variantTag} <em>[variant not in database]</em>
-        </div>
-      )}
+  <div style={{ marginTop: '1em' }}>
+    <h3>Variant</h3>
+    <VariantTagCard data={variantTag} />
```
conradarcturus, line 84: There isn't enough space for a card here -- but if you add a HoverableObjectName you will see the VariantCardTag on hover.
```suggestion
    <HoverableObjectName object={variantTag} />
```

In order for this to work you just need to update HoverableObject to handle variant cards, here: https://github.com/Translation-Commons/lang-nav/blob/master/src/views/common/HoverableObject.tsx#L24-L38

src/views/locale/LocaleDetails.tsx
```diff
@@ -78,11 +79,11 @@ const LocaleDefinitionSection: React.FC<{ locale: LocaleData }> = ({ locale }) =
         </div>
       )}
       {variantTag && (
-        <div>
-          <label>Variant:</label>
-          {variantTag} <em>[variant not in database]</em>
-        </div>
-      )}
+  <div style={{ marginTop: '1em' }}>
```
conradarcturus, line 82: You shouldn't need the Margin here. Also there is something wrong with the indentation -- you should install the Prettier and es-lint plugins on VSCode.

src/types/PageParamTypes.tsx
```diff
@@ -7,7 +7,7 @@ export enum ObjectType {
   Census = 'Census',
   Territory = 'Territory',
   WritingSystem = 'Writing System',
-  // VariantTag = 'Variant Tag',
+  VariantTag = 'Variant Tag',
```
conradarcturus, line 10: yay!

src/types/DataTypes.tsx
```diff
@@ -16,7 +16,13 @@ export interface ObjectBase {
   names: string[];
 }
 
-export type ObjectData = CensusData | LanguageData | LocaleData | TerritoryData | WritingSystemData;
+export type ObjectData =
```
conradarcturus, line 19: this all lgtm, thanks!

src/data/CoreData.tsx
```diff
@@ -1,5 +1,9 @@
 import { useState } from 'react';
 
+import { VariantTagData } from '../types/DataTypes';
```
conradarcturus, line 3: Add this to the DataTypes import below

src/data/CoreData.tsx
```diff
@@ -116,6 +121,8 @@ export function useCoreData(): {
     addGlottologLanguages(languagesBySchema, glottologImport || [], manualGlottocodeToISO || {});
     addCLDRLanguageSchema(languagesBySchema, cldrAliases || []);
     addIANAVariantLocales(languagesBySchema, locales, ianaVariants);
+    const variantTags = convertToVariantTagData(ianaVariants || []);
+    connectVariantTags(variantTags, languagesBySchema.Inclusive, locales);
```
conradarcturus, line 125: You'll want to use CLDR here. "Inclusive" prefers 3-letter language codes but "CLDR" uses 2-letter language codes and is to how the IANA catalog is set up.
```suggestion
    connectVariantTags(variantTags, languagesBySchema.CLDR, locales);
```

src/data/CoreData.tsx
```diff
@@ -39,14 +43,15 @@ import {
...
-import { loadIANAVariants, addIANAVariantLocales } from './IANAData';
+import { loadIANAVariants, addIANAVariantLocales, convertToVariantTagData } from './IANAData';
 
 export type CoreData = {
   censuses: Record<CensusID, CensusData>;
   languagesBySchema: LanguagesBySchema;
   locales: Record<BCP47LocaleCode, LocaleData>;
   territories: Record<TerritoryCode, TerritoryData>;
   writingSystems: Record<ScriptCode, WritingSystemData>;
+  variantTags: VariantTagData[];
```
conradarcturus, line 54: Good! You'll want to make a setter function like `const [locales, setLocales] = useState<Record<BCP47LocaleCode, LocaleData>>({});` below

You'll want to use the datatype `Record<VariantIANATag, VariantTagData>` -- you can always cast this back to an array using `Object.values(...)`

In order to make a dictionary, you can just do `Object.fromEntries(variantDataArray.map(v => [v.ID, v]))`

src/data/CoreData.tsx
```diff
@@ -138,6 +145,7 @@ export function useCoreData(): {
       locales,
       territories,
       writingSystems,
+      variantTags,
```
conradarcturus, line 148: You'll see to instantiate variant Tags with that aformentioned setter function. Then around line 135 you see the function `setWritingSystems(writingSystems);` -- you'll set it there in the inner function.

src/data/IANAData.tsx
```diff
@@ -18,6 +18,22 @@ export async function loadIANAVariants(): Promise<IANAVariantData[] | void> {
           .catch((err) => console.error('Error loading TSV:', err));
 }
 
+export function convertToVariantTagData(
```
conradarcturus, line 21: Instead of having this convert function, why not just update the IANAVariantData type & import?

src/data/connectVariantTags.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    // Link to LocaleData
+    variant.locales = [];
+
+    for (const locale of Object.values(locales)) {
+      if (locale.variantTagID === variant.ID) {
+        variant.locales.push(locale);
+        locale.variantTag = variant;
+      }
+    }
```
conradarcturus, line 28: This is assymptotically very expensive. You are iterating over the entire list of locales (10,000s of entries) for every variant tag.

Rather, it would be better to make a separate loop outside of the for loop you used on line 9. That can iterate over all locales and if `locale.variantTagID != null` then it can query the VariantTag dictionary mentioned above.

conradarcturus (changes requested):
Thanks for the update on your progress. Here are some helpful tips.

Also, you may see there are rebasing conflicts. You should refresh your copy of master and rebase this branch on master then force-push the update.

conradarcturus (approved):
I made a number of updates so we can land this code -- it's been cooking for awhile, it would be nice to merge the changes.

If you enjoyed working on this, there are many followup tasks: https://github.com/Translation-Commons/lang-nav/issues/143

PR #121, opened 2025-07-24
new commits

wizardsimms:


conradarcturus:
@wizardsimms Please don't use public pull requests to rebase your branch. All PRs need a review from the code-owners so a PR like this will email us all. Also, since it isn't actually a functional change it dilutes the PR history by adding non-functional PRs.

You can do this yourself on your devserver. I'll walk you through that in our meeting later today.

PR #195, opened 2025-08-19
Add variant tag population calculations and displays

wizardsimms:
Implemented two figures for each variant: a cited population derived from available locale data, and an upper-bound population estimate based on the total population of the variant’s associated prefix or language (for example, roh_rumgr could have as many users as roh in general). These additions make sorting and comparing variants by population more meaningful even when direct variant-specific population data is missing.

### DIFF
--- a/src/views/common/ObjectField.tsx
+++ b/src/views/common/ObjectField.tsx
@@ -99,7 +99,24 @@ export function getObjectPopulation(
       );
     case ObjectType.Territory:
       return object.population;
-    case ObjectType.VariantTag:
-      return object.languages.reduce((sum, lang) => sum + (lang.populationCited || 0), 0);
+    case ObjectType.VariantTag: {
+      // For variant tags, attempt to return a cited population first (based on locale data),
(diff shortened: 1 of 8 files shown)


src/views/common/ObjectField.tsx
```diff
@@ -88,10 +88,26 @@ export function getObjectPopulation(object: ObjectData): number {
...
+        return cited;
+      }
+      const upper = variant.populationUpperBound;
+      if (typeof upper === 'number' && upper > 0) {
+        return upper;
+      }
+      // Fallback: sum the populationCited values from associated languages
+      return (variant.languages || []).reduce((sum: number, lang: any) => {
+        return sum + (lang.populationCited || 0);
+      }, 0);
```
conradarcturus, line 109: Much of this code is wrong. There is no reason to do `const variant = object as any;` and that's causing further problems.

conradarcturus (changes requested):
Fix build errors & include a test plan with screenshots in the PR description -- otherwise I cannot evaluate the code and how it completes the task.

PR #203, opened 2025-09-11
Table: add export button

wizardsimms:
Screenshot 2025-09-11 at 1 42 45 PM

### DIFF
--- a/test-coverage.md
+++ b/test-coverage.md
@@ -2,11 +2,11 @@
 
 | Path                                     |   Lines | Branches |   Funcs |   Stmts |
 | ---------------------------------------- | ------: | ------: | ------: | ------: |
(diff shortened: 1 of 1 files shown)


src/views/common/table/ObjectTable.tsx
```diff
@@ -72,19 +129,105 @@ function ObjectTable<T extends ObjectData>({
...
+                type="checkbox"
+                checked={visibleColumns[column.key] || false}
+                onChange={() => toggleColumn(column.key)}
+                style={{ marginRight: '0.5rem' }}
+              />
+              {column.label ?? column.key}
+            </label>
+          ))}
+        </div>
+      </details>
```
conradarcturus: Looks like you kept the old table column selector component when you merged, you don't need it since that code has been moved to `TableColumnSelector`.

```suggestion
```

PR #204, opened 2025-09-11
Locales: Add CLDR support per locale

wizardsimms:
Add CLDR locale support by ingesting CLDR data for locales (e.g., en_US, en_GB) and displaying coverage indicators in LocaleTable and LocaleDetails, similar to how language-level CLDR support is shown.

### DIFF
--- a/public/data/unicode/cldrLocales.json
+++ b/public/data/unicode/cldrLocales.json
@@ -1,5 +1,5 @@
 {
-  "release": "0.0.0",
+  "release": "43.0.0",
   "generatedAt": "",
   "locales": []
 }
--- a/scripts/ingest/build-cldr-locales.ts
+++ b/scripts/ingest/build-cldr-locales.ts
@@ -18,7 +18,14 @@
 
 import fs from 'node:fs/promises';
 import path from 'node:path';
-import territoryInfo from 'cldr-core/supplemental/territoryInfo.json';
+
+const CLDR_RELEASE = '43.0.0';
+const CLDR_CORE_BASE = `https://cdn.jsdelivr.net/npm/cldr-core@${CLDR_RELEASE}`;
+const CHARTS_TSV_BASE = `https://raw.githubusercontent.com/unicode-org/cldr-staging/main/docs/charts/${CLDR_RELEASE.replace(
+  /\.0\.0$/,
+  '',
+)}/tsv`;
+const CLDR_REPO_RAW_BASE = 'https://raw.githubusercontent.com/unicode-org/cldr/main/common/main';
 
 const OUTPUT_FILE = path.join(process.cwd(), 'public/data/unicode/cldrLocales.json');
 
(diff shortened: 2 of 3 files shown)


scripts/ingest/build-cldr-locales.ts
```diff
@@ -0,0 +1,197 @@
...
+// both the charts path and the cldr-core package version.
+const CLDR_RELEASE = '47.0.0';
+
+// Base URLs (pinned to the release above)
+const CLDR_CORE_BASE = `https://cdn.jsdelivr.net/npm/cldr-core@${CLDR_RELEASE}`;
+const CHARTS_TSV_BASE = `https://raw.githubusercontent.com/unicode-org/cldr-staging/main/docs/charts/${CLDR_RELEASE.replace(
+  /\.0\.0$/,
+  '',
+)}/tsv`;
+const CLDR_REPO_RAW_BASE = 'https://raw.githubusercontent.com/unicode-org/cldr/main/common/main';
```
conradarcturus, line 32: I'm not certain I want to ingest the data -- what were the tradeoffs that you considered when you decided to use these URLs?

There is the CLDR Json repository that we already have access to data in. You can see how we use it in files like `UnicodeData.tsx` where we access the data with imports like `import territoryInfo from 'cldr-core/supplemental/territoryInfo.json';`

scripts/ingest/build-cldr-locales.ts
```diff
@@ -0,0 +1,197 @@
+/**
+ * Build script to generate a list of CLDR locale support objects.
```
conradarcturus, line 2: Thanks for including the large comment block.

public/data/unicode/cldrLocales.json
```diff
@@ -0,0 +1,5 @@
+{
+  "release": "0.0.0",
+  "generatedAt": "",
+  "locales": []
```
conradarcturus, line 4: This output file is empty -- is there supposed to be data in it?

src/types/CLDRLocaleTypes.tsx
```diff
@@ -0,0 +1,63 @@
...
+  locale: string;
+  /** Language code, e.g. "en" */
+  language: string;
+  /** Region code, e.g. "US" (optional) */
+  region?: string;
+  /** Script code, e.g. "Latn" (optional) */
+  script?: string;
+
+  /** Tier of locale support: core, modern or full */
+  tier: 'core' | 'modern' | 'full';
```
conradarcturus, line 17: This doesn't look like the CLDR levels that I am used to. I'm used to seeing "core", "basic", "moderate", "modern". Where do these tiers come from?

conradarcturus (changes requested):
There appears to be a lot of good work in this change. However, I won't give a full review until you post a test plan. Please add how you tested it, how you run the new command you added, and include screenshots.

PR #212, opened 2025-09-18
Feat/vitality meter 85

### LINKED ISSUE #85: UX: Replace Language Card "Vitality" text with Meter display
If you see [UNESCO's World Atlas](https://en.wal.unesco.org/), it uses meter bars to show the language vitality
Image

Let's change our language card display to follow a similar design pattern, replacing what is show for "Vitality" today.
Image

See the HTML <meter> component reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meter

The current vitality displayed comes from the variable `vitalityEth2013` You can see expanded explanations here: https://www.ethnologue.com/methodology/#language-status

abdulkalam556:
## Summary
Replaces the raw Ethnologue vitality display with an accessible HTML `<meter>` on language cards. On the details view, shows both the original text and the meter (per spec).

Closes #85.

## Implementation details
- Added `LanguageVitalityMeter` in `src/views/language/`:
  - Accepts 2013 EGIDS-like `value` (number or string, e.g., "6b Threatened", 7, "National").
  - Normalizes to base EGIDS 0..10 (collapsing 6a/6b → 6 and 8a/8b → 8).
  - Inverts to meter range 0..9 with `min=0`, `max=9`, `optimum=9`.
  - Skips "International" (0) as N/A.
  - When null/empty → `<Deemphasized>Data not available</Deemphasized>`.

- `LanguageCard.tsx`: replaced `{vitalityEth2013}` with `<LanguageVitalityMeter value={vitalityEth2013} />`.
- `LanguageDetails.tsx`: shows the raw `vitalityEth2013` value and `<LanguageVitalityMeter value={vitalityEth2013} showLabel />`.

## Screenshots
Screenshot 2025-09-17 at 2 26 13 PM

## Notes:
in the deatils, the for the Vitality & Viability section, there is Ethnologue Vitality (2025) looks like the values of these are different from Ethnologue Vitality (2023) so couldn't use the <LanguageVitalityMeter value={vitalityEth2013} />
please see the below image for more.
Screenshot 2025-09-18 at 12 30 32 PM

### DIFF
--- a/src/views/language/LanguageVitalityMeter.tsx
+++ b/src/views/language/LanguageVitalityMeter.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+
+import Deemphasized from '../../generic/Deemphasized';
+
+interface Props {
+  value?: string;
+}
+
+const LanguageVitalityMeter: React.FC<Props> = ({ value }) => {
+  if (!value) {
+    return <Deemphasized>Data not available</Deemphasized>;
+  }
+
+  // Map Ethnologue vitality levels to meter values (inverted scale)
+  // Ethnologue scale: 1=National, 2=Regional, 3=Trade, 4=Educational, 5=Written, 6=Threatened, 7=Shifting, 8=Moribund, 9=Nearly Extinct, 10=Extinct
+  // We want: 9=National, 8=Regional, 7=Trade, 6=Educational, 5=Written, 4=Threatened, 3=Shifting, 2=Moribund, 1=Nearly Extinct, 0=Extinct
+  const getMeterValue = (vitality: string): number => {
+    switch (vitality.toLowerCase()) {
+      case 'national':
+        return 9;
+      case 'regional':
+        return 8;
+      case 'trade':
+        return 7;
+      case 'educational':
+        return 6;
+      case 'written':
+        return 5;
+      case 'threatened':
+        return 4;
+      case 'shifting':
+        return 3;
+      case 'moribund':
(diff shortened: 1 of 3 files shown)


src/views/language/LanguageVitalityMeter.tsx
```diff
@@ -0,0 +1,58 @@
...
+      default:
+        // For any unknown values, return a neutral value
+        return 5;
+    }
+  };
+
+  const meterValue = getMeterValue(value);
+
+  return (
+    <meter
```
conradarcturus, line 48: Instead of using the title, use our built-in hoverable

```suggestion
    <Hoverable hoverContent={value}>
      <meter
```

src/views/language/LanguageVitalityMeter.tsx
```diff
@@ -0,0 +1,58 @@
...
+      min={0}
+      max={9}
+      value={meterValue}
+      title={`Vitality: ${value}`}
+      style={{ width: '100%' }}
+    />
+  );
```
conradarcturus, line 55: Make a minWidth to make the Details page view work better without a flexbox.

```suggestion
        style={{ width: '100%', minWidth: '8em' }}
      />
    </Hoverable>
  );
```

conradarcturus (changes requested):
Thanks for your patience, I'm feeling better now. I left some style suggestions, with htem you should be good to land the change.

PR #220, opened 2025-09-25
feat: implement vitality metascore calculation and integrate (#213)

### LINKED ISSUE #213: Vitality: Make meta-score for Vitality
# Context

We currently have 3 different vitality estimates. They don't always agree and they don't label every single language.

* ISO (2025) Living or Constructed or Historical or Extinct
* Ethnologue (2013) [explanation](https://www.ethnologue.com/methodology/#language-status)
* Ethnologue (2025) Institutional or Stable or Endangered or Extinct

# Task
When we show the vitality as a meter we should show a meta-score. It will be challenging to reconcile the 3 input scores.

To start, you'll want to prepare an algorithm to make a vitality score. One possible approach is:
* Convert Ethnologue 2013 to 9 point scale as already done
* Convert Ethnologue 2025 to 9 point scale too, with values Institutional = 9, Stable = 6, Endangered = 3 and Extinct = 0 
* If 2013 & 2025 values exist, return the average of the 2
* If only 1 ethnologue value exists, return that
* If neither ethnologue value exists, use ISO scale, Living = 9, Constructed = 3, Historical = 1, Extinct = 0

You'll probably want that in a separate file eg. `LanguageVitalityComputation.tsx` with functions mapping the raw input/enums to 0-9 scores. You'd want to update the `LanguageVitalityMeter.tsx` component to use this score to determine the length. With this algorithm, you'll want to show the Metascore in LanguageCard and the 4 vitality scores in the LanguageDetails. For example, Esperanto is Vigorous in 2013 (4) but Institutional in 2025 (9) so the Metascore is 6.5. You can show separate scores for the individual measures. Make them all hoverable to explain where it comes from (show the calculation for the metascore in the hoverable for the metascore). So Language Details will look like this:

Vitality Metascore: <------     >
ISO Vitality / Status: Constructed <---        >
Ethnologue (2013): Vigorous <----       >
Ethnologue (2025): Institutional <--------->

# Value overlap

This is a screenshot showing the number of languages with overlapping labels from the original master spreadsheet we prepared earlier this year. This shows how some languages have different vitality estimates for different places.

Image

abdulkalam556:
### Summary
Implements the **Vitality Metascore** system to reconcile ISO, Ethnologue 2013, and Ethnologue 2025 vitality measures.

### Changes
- **`LanguageVitalityComputation.tsx`**
  - `getEthnologue2013Score()`: maps values to 0–9 scale  
  - `getEthnologue2025Score()`: maps values to 0–9 scale (Institutional=9, Stable=6, Endangered=3, Extinct=0)  
  - `getISOScore()`: maps ISO values to 0–9 scale (Living=9, Constructed=3, Historical=1, Extinct=0)  
  - `computeVitalityMetascore()`:  
    - If both Ethnologue values exist → average  
    - If only one exists → use that  
    - If none exist → fallback to ISO scale  
  - `getAllVitalityScores()`: returns all vitality scores for display  

- **`LanguageVitalityMeter.tsx`**
  - Accepts a `LanguageData` object  
  - Uses the metascore computation  
  - Shows hover explanation with calculation details  

- **`LanguageCard.tsx`**
  - Displays "Vitality Metascore" instead of plain "Vitality"  
  - Uses new metascore system  

- **`LanguageDetailsVitalityAndViability.tsx`**
  - Displays:  
    - Vitality Metascore (meter + hover details)  
    - ISO Vitality / Status  
    - Ethnologue (2013)  
    - Ethnologue (2025)  
  - All values shown with meters and hover explanations  

### Screenshots
this is new look of languageCard:
Screenshot 2025-09-25 at 3 44 37 PM

below are the images for the languageCardDetails:
Screenshot 2025-09-25 at 3 47 32 PM
Screenshot 2025-09-25 at 3 47 43 PM
Screenshot 2025-09-25 at 3 47 57 PM
Screenshot 2025-09-25 at 3 48 03 PM

### Issue
Closes #213

### DIFF
--- a/src/views/language/LanguageCard.tsx
+++ b/src/views/language/LanguageCard.tsx
@@ -20,7 +20,7 @@ interface Props {
 const LanguageCard: React.FC<Props> = ({ lang, includeRelations }) => {
   const { updatePageParams } = usePageParams();
   const sortFunction = getSortFunction();
-  const { ID, locales, modality, populationEstimate, vitalityEth2013 } = lang;
+  const { ID, locales, modality, populationEstimate } = lang;
   const countryLocales = uniqueBy(
     locales.filter((l) => l.territory?.scope === TerritoryScope.Country).sort(sortFunction),
     (l) => l.territoryCode,
@@ -48,8 +48,8 @@ const LanguageCard: React.FC<Props> = ({ lang, includeRelations }) => {
         </div>
       )}
       <div>
-        <h4>Vitality</h4>
-        <LanguageVitalityMeter value={vitalityEth2013} />
+        <h4>Vitality Metascore</h4>
+        <LanguageVitalityMeter lang={lang} />
       </div>
 
       {includeRelations && countryLocales.length > 0 && (
--- a/src/views/language/LanguageCard.tsx
+++ b/src/views/language/LanguageCard.tsx
@@ -11,6 +11,7 @@ import ObjectTitle from '../common/ObjectTitle';
 import PopulationWarning from '../common/PopulationWarning';
 
 import LanguageVitalityMeter from './LanguageVitalityMeter';
+import { VitalityMeterType } from './LanguageVitalityComputation';
 
 interface Props {
   lang: LanguageData;
@@ -49,7 +50,7 @@ const LanguageCard: React.FC<Props> = ({ lang, includeRelations }) => {
       )}
       <div>
         <h4>Vitality</h4>
-        <LanguageVitalityMeter lang={lang} />
+        <LanguageVitalityMeter lang={lang} type={VitalityMeterType.Metascore} />
       </div>
 
       {includeRelations && countryLocales.length > 0 && (
--- a/src/views/language/LanguageCard.tsx
+++ b/src/views/language/LanguageCard.tsx
@@ -10,8 +10,8 @@ import HoverableObjectName from '../common/HoverableObjectName';
 import ObjectTitle from '../common/ObjectTitle';
 import PopulationWarning from '../common/PopulationWarning';
 
-import LanguageVitalityMeter from './LanguageVitalityMeter';
 import { VitalityMeterType } from './LanguageVitalityComputation';
+import LanguageVitalityMeter from './LanguageVitalityMeter';
 
 interface Props {
   lang: LanguageData;
--- a/src/views/language/LanguageCard.tsx
+++ b/src/views/language/LanguageCard.tsx
@@ -48,7 +48,7 @@ const LanguageCard: React.FC<Props> = ({ lang, includeRelations }) => {
         </div>
       )}
       <div>
-        <h4>Vitality Metascore</h4>
+        <h4>Vitality</h4>
         <LanguageVitalityMeter lang={lang} />
       </div>
 
--- a/src/views/language/LanguageDetailsVitalityAndViability.tsx
+++ b/src/views/language/LanguageDetailsVitalityAndViability.tsx
@@ -1,34 +1,76 @@
 import React from 'react';
 
+import Hoverable from '../../generic/Hoverable';
 import { LanguageData } from '../../types/LanguageTypes';
 import { CLDRCoverageText, ICUSupportStatus } from '../common/CLDRCoverageInfo';
 import DetailsField from '../common/details/DetailsField';
 import DetailsSection from '../common/details/DetailsSection';
 import ObjectWikipediaInfo from '../common/ObjectWikipediaInfo';
 
+import { getAllVitalityScores } from './LanguageVitalityComputation';
 import LanguageVitalityMeter from './LanguageVitalityMeter';
 
+// Component for individual vitality meters with hover explanations
+const VitalityMeter: React.FC<{
+  score: number | null;
+  value: string | null;
+  title: string;
+  explanation: string;
+}> = ({ score, value, title, explanation }) => {
+  if (score === null || value === null) {
+    return null;
+  }
+
+  return (
+    <DetailsField title={title}>
+      <span>{value} </span>
+      <Hoverable hoverContent={`${title}: ${score} (${explanation})`}>
+        <meter
+          min={0}
+          low={3}
+          high={7}
+          optimum={8}
+          max={9}
+          value={score}
+          title={`${title}: ${score}`}
+          style={{ width: '100%', minWidth: '8em' }}
+        />
+      </Hoverable>
+    </DetailsField>
+  );
+};
+
 const LanguageDetailsVitalityAndViability: React.FC<{ lang: LanguageData }> = ({ lang }) => {
-  const {
-    vitalityISO,
-    vitalityEth2013,
-    vitalityEth2025,
-    viabilityConfidence,
-    viabilityExplanation,
-  } = lang;
+  const { viabilityConfidence, viabilityExplanation } = lang;
+
+  const vitalityScores = getAllVitalityScores(lang);
 
   return (
     <DetailsSection title="Vitality & Viability">
-      {vitalityISO && <DetailsField title="ISO Vitality / Status:">{vitalityISO}</DetailsField>}
-      {vitalityEth2013 && (
-        <DetailsField title="Ethnologue Vitality (2013):">
-          <span>{vitalityEth2013} </span>
-          <LanguageVitalityMeter value={vitalityEth2013} />
-        </DetailsField>
-      )}
-      {vitalityEth2025 && (
-        <DetailsField title="Ethnologue Vitality (2025):">{vitalityEth2025}</DetailsField>
-      )}
+      <DetailsField title="Vitality Metascore:">
+        <LanguageVitalityMeter lang={lang} />
+      </DetailsField>
+
+      <VitalityMeter
+        score={vitalityScores.iso.score}
+        value={vitalityScores.iso.value}
+        title="ISO Vitality / Status:"
+        explanation={`ISO scale: Living=9, Constructed=3, Historical=1, Extinct=0`}
+      />
+
+      <VitalityMeter
+        score={vitalityScores.eth2013.score}
+        value={vitalityScores.eth2013.value}
+        title="Ethnologue (2013):"
+        explanation={`Ethnologue 2013 scale: National=9, Regional=8, Trade=7, Educational=6, Written=5, Threatened=4, Shifting=3, Moribund=2, Nearly Extinct=1, Extinct=0`}
+      />
+
+      <VitalityMeter
+        score={vitalityScores.eth2025.score}
+        value={vitalityScores.eth2025.value}
+        title="Ethnologue (2025):"
+        explanation={`Ethnologue 2025 scale: Institutional=9, Stable=6, Endangered=3, Extinct=0`}
+      />
       <DetailsField title="Should use in World Atlas:">
         {viabilityConfidence} ... {viabilityExplanation}
       </DetailsField>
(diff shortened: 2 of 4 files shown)


src/views/language/LanguageVitalityMeter.tsx
```diff
@@ -2,59 +2,33 @@ import React from 'react';
...
-        low={3} // Shifting
-        high={7} // Trade
-        optimum={8} // Regional
-        max={9} // National
-        value={meterValue}
-        title={`Vitality: ${value}`}
+        low={3} // Shifting/Endangered
+        high={7} // Trade/Stable
+        optimum={8} // Regional/Institutional
+        max={9} // National/Living
```
conradarcturus, line 29: I love the comments, just fixing an error and explaining the color better
```suggestion
        low={3} // Shifting/Endangered -- below this, the meter is colored red
        high={7} // Trade/Stable -- below this, the meter is colored yellow
        optimum={8} // Regional -- tells the renderer that high and above is green
        max={9} // National/Institutional/Living
```

src/views/language/LanguageVitalityMeter.tsx
```diff
@@ -2,59 +2,33 @@ import React from 'react';
...
+import { computeVitalityMetascore } from './LanguageVitalityComputation';
 
 interface Props {
-  value?: string;
+  lang: LanguageData;
 }
 
-const LanguageVitalityMeter: React.FC<Props> = ({ value }) => {
-  if (!value) {
+const LanguageVitalityMeter: React.FC<Props> = ({ lang }) => {
```
conradarcturus, line 13: Instead of specializing the component for the Metascore -- can you pass in an enum of the 4 vitality types and it will show a meter with the appropriate values and hover content?

src/views/language/LanguageVitalityComputation.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+
+/**
+ * Computes the vitality metascore for a language using the algorithm:
+ * 1. If both Ethnologue 2013 & 2025 values exist, return the average
+ * 2. If only one Ethnologue value exists, return that
+ * 3. If neither Ethnologue value exists, use ISO scale
+ */
+export function computeVitalityMetascore(lang: LanguageData): {
```
conradarcturus, line 85: I like the method & explanation -- very effective.

src/views/language/LanguageVitalityComputation.tsx
```diff
@@ -0,0 +1,155 @@
...
+  score: number | null;
+  explanation: string;
```
conradarcturus, line 87: (maybe, depends on your other changes). Instead of a `string` can you return a `ReactNode`? That way you can add light formatting -- otherwise everything is just one 1 single line. For instance, for the Metascore you can include the main score and then the constituent scores indented by `2em` or something.

src/views/language/LanguageVitalityComputation.tsx
```diff
@@ -0,0 +1,155 @@
...
+    case 'threatened':
+      return 4;
+    case 'shifting':
+      return 3;
+    case 'moribund':
+      return 2;
+    case 'nearly extinct':
+      return 1;
+    case 'extinct':
+      return 0;
```
conradarcturus, line 31: Looks like you are missing some values. Check the spreadsheet lookup results. You can add extra cases to be defensive.

Image
```suggestion
    case 'national':
      return 9;
    case 'provincial':
    case 'regional':
      return 8;
    case 'trade':
    case 'wider communication':
      return 7;
    case 'educational':
      return 6;
    case 'written':
    case 'developing':
      return 5;
    case 'vigorous':
    case 'threatened':
      return 4;
    case 'shifting':
      return 3;
    case 'moribund':
    case 'nearly extinct':
      return 2;
    case 'dormant':
      return 1;
    case 'extinct':
      return 0;
```

src/views/language/LanguageVitalityComputation.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  switch (vitality.toLowerCase()) {
+    case 'institutional':
+      return 9;
+    case 'stable':
+      return 6;
+    case 'endangered':
+      return 3;
+    case 'extinct':
+      return 0;
```
conradarcturus, line 52: Missing numbers -- you'll note if you test this that the 2025 numbers never actually appear.

```suggestion
    case '1 institutional':
      return 9;
    case '2 stable':
      return 6;
    case '3 endangered':
      return 3;
    case '4 extinct':
      return 0;
```

src/views/language/LanguageVitalityComputation.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const sources: string[] = [];
+  let explanation = '';
+
+  if (eth2013Score !== null && eth2025Score !== null) {
+    // Both Ethnologue values exist - return average
+    const average = (eth2013Score + eth2025Score) / 2;
+    sources.push('Ethnologue 2013', 'Ethnologue 2025');
+    explanation = `Average of Ethnologue 2013 (${eth2013Score}) and Ethnologue 2025 (${eth2025Score}) = ${average}`;
+    return { score: average, explanation, sources };
+  }
```
conradarcturus, line 103: It doesn't look like you are actually using the intermediates. Also it's probably better to use the text in the explanation rather than the scores.

You can also consider adding a link to ethnologue https://www.ethnologue.com/methodology/#languageStatus

```suggestion
  if (eth2013Score !== null && eth2025Score !== null) {
    // Both Ethnologue values exist - return average
    return { 
      score: (eth2013Score + eth2025Score) / 2, 
      explanation: <>Average of 2 ethnologue scores:
        <div style={{marginLeft: "2em"}}>Ethnologue 2013 ({lang.vitalityEth2013})</div>
        <div style={{marginLeft: "2em"}}>Ethnologue 2025 ({lang.vitalityEth2025})</div>
      </>,
    };
  }
```

src/views/language/LanguageVitalityComputation.tsx
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Computes the vitality metascore for a language using the algorithm:
+ * 1. If both Ethnologue 2013 & 2025 values exist, return the average
+ * 2. If only one Ethnologue value exists, return that
+ * 3. If neither Ethnologue value exists, use ISO scale
+ */
+export function computeVitalityMetascore(lang: LanguageData): {
+  score: number | null;
+  explanation: string;
+  sources: string[];
```
conradarcturus, line 88: You don't seem to use it. I think a rich-text explanation is all that's necessary anyway.

```suggestion
```

src/views/language/LanguageDetailsVitalityAndViability.tsx
```diff
@@ -1,34 +1,76 @@
...
 import { CLDRCoverageText, ICUSupportStatus } from '../common/CLDRCoverageInfo';
 import DetailsField from '../common/details/DetailsField';
 import DetailsSection from '../common/details/DetailsSection';
 import ObjectWikipediaInfo from '../common/ObjectWikipediaInfo';
 
+import { getAllVitalityScores } from './LanguageVitalityComputation';
 import LanguageVitalityMeter from './LanguageVitalityMeter';
 
+// Component for individual vitality meters with hover explanations
+const VitalityMeter: React.FC<{
```
conradarcturus, line 14: As noted below you should probably use 1 central meter component rather than calling it twice.

src/views/language/LanguageDetailsVitalityAndViability.tsx
```diff
@@ -1,34 +1,76 @@
...
+        value={vitalityScores.iso.value}
+        title="ISO Vitality / Status:"
+        explanation={`ISO scale: Living=9, Constructed=3, Historical=1, Extinct=0`}
+      />
+
+      <VitalityMeter
+        score={vitalityScores.eth2013.score}
+        value={vitalityScores.eth2013.value}
+        title="Ethnologue (2013):"
+        explanation={`Ethnologue 2013 scale: National=9, Regional=8, Trade=7, Educational=6, Written=5, Threatened=4, Shifting=3, Moribund=2, Nearly Extinct=1, Extinct=0`}
```
conradarcturus, line 65: This is pretty length. It may be better to use a rich-text component. Also tbh I'm not sure this breakdown is necessary, I'd almost direct people to a readme or the raw code itself.

A question I ask myself (but I leave to you to decide, and not necessarily in this change) is how transparent we should be on what the scores are -- are they meaningful numbers we would want people to export when downloading our dataset? Or are they just a utility to render the meter progress?

src/views/language/LanguageDetailsVitalityAndViability.tsx
```diff
@@ -1,34 +1,76 @@
...
+      <span>{value} </span>
+      <Hoverable hoverContent={`${title}: ${score} (${explanation})`}>
+        <meter
+          min={0}
+          low={3}
+          high={7}
+          optimum={8}
+          max={9}
+          value={score}
+          title={`${title}: ${score}`}
```
conradarcturus, line 35: I forgot to say this before, we don't need this title component since we provide it with our own custom Hoverable.

src/views/language/LanguageVitalityMeter.tsx
```diff
@@ -2,59 +2,33 @@ import React from 'react';
...
-        optimum={8} // Regional
-        max={9} // National
-        value={meterValue}
-        title={`Vitality: ${value}`}
+        low={3} // Shifting/Endangered
+        high={7} // Trade/Stable
+        optimum={8} // Regional/Institutional
+        max={9} // National/Living
+        value={metascore.score}
+        title={`Vitality Metascore: ${metascore.score.toFixed(1)}`}
```
conradarcturus, line 31: As noted above, I don't think a title is necessary because we provide a hovercard with the hoverable.
```suggestion
```

src/views/language/LanguageCard.tsx
```diff
@@ -48,8 +48,8 @@ const LanguageCard: React.FC<Props> = ({ lang, includeRelations }) => {
         </div>
       )}
       <div>
-        <h4>Vitality</h4>
-        <LanguageVitalityMeter value={vitalityEth2013} />
+        <h4>Vitality Metascore</h4>
```
conradarcturus, line 51: You can just call it vitality.

```suggestion
        <h4>Vitality</h4>
```

conradarcturus (changes requested):
This is looking really good! Functionally it is almost there, most of my feedback is just on styling code and visuals. I appreciate the fast iteration on this. I'm really happy to have you on the team.

conradarcturus:
Thanks for the updates -- it looks like there are some build errors though.

src/views/language/LanguageVitalityMeter.tsx
abdulkalam556: made LanguageVitalityMeter the central component

src/views/language/LanguageDetailsVitalityAndViability.tsx
abdulkalam556: I felt the same, probably we need to show whole scale instead beside score in () I started showing the what the score meant. eg: 9 (National) in 2013 ethnologue

src/views/language/LanguageDetailsVitalityAndViability.tsx
abdulkalam556: removed Title

conradarcturus (changes requested):
You still need to make sure the switch statements use all of the correct case values. Also the hoverText is too verbose and can be simplified.

PR #239, opened 2025-10-03
Add Vitality columns to LanguageTable

abdulkalam556:
- Add 4 new vitality columns: Meta, ISO, Eth 2013, Eth 2025
- Use wifi icons (WifiHigh/Wifi/WifiLow/WifiOff) for visual indicators
- Color code scores: green (strong), yellow (medium), orange (low), red (extinct)
- Show numeric scores (0-9) with hover explanations
- Group columns under 'Vitality' category
- Metascore column visible by default, others hidden initially"

Screenshot 2025-10-03 at 12 05 02 PM

### DIFF
--- a/src/views/language/LanguageTable.tsx
+++ b/src/views/language/LanguageTable.tsx
@@ -1,4 +1,4 @@
-import { InfoIcon, TriangleAlertIcon } from 'lucide-react';
+import { InfoIcon, TriangleAlertIcon, Wifi, WifiHigh, WifiLow, WifiOff } from 'lucide-react';
 import React, { ReactNode } from 'react';
 
 import { getUniqueTerritoriesForLanguage } from '../../controls/sort';
@@ -7,6 +7,12 @@ import Hoverable from '../../generic/Hoverable';
 import HoverableEnumeration from '../../generic/HoverableEnumeration';
 import { LanguageData, LanguageField } from '../../types/LanguageTypes';
 import { SortBy } from '../../types/PageParamTypes';
+import {
+  computeVitalityMetascore,
+  getAllVitalityScores,
+  getVitalityExplanation,
+  VitalityMeterType,
+} from './LanguageVitalityComputation';
 import { CLDRCoverageText, ICUSupportStatus } from '../common/CLDRCoverageInfo';
 import HoverableObjectName from '../common/HoverableObjectName';
 import ObjectWikipediaInfo from '../common/ObjectWikipediaInfo';
@@ -16,6 +22,38 @@ import ObjectTable from '../common/table/ObjectTable';
 
 const LanguageTable: React.FC = () => {
   const { languagesInSelectedSource } = useDataContext();
+
+  // Helper functions for vitality display
+  function scoreBucket(score: number | null): 0 | 1 | 2 | 3 {
+    if (score === null) return 0;
+    if (score >= 7) return 3;     // strong
+    if (score >= 4) return 2;     // medium
+    if (score >= 1) return 1;     // low
+    return 0;                     // none/extinct
+  }
+
+  function bucketIcon(bucket: 0 | 1 | 2 | 3) {
+    switch (bucket) {
+      case 3: return <WifiHigh size="1em" />;
+      case 2: return <Wifi size="1em" />;
+      case 1: return <WifiLow size="1em" />;
+      default: return <WifiOff size="1em" />;
+    }
+  }
+
+  function bucketColor(bucket: 0 | 1 | 2 | 3): string {
+    switch (bucket) {
+      case 3: return 'var(--color-text-green)';
+      case 2: return 'var(--color-text-yellow)';
+      case 1: return 'var(--color-text-orange)';
+      default: return 'var(--color-text-red)';
+    }
+  }
+
+  function scoreLabel(score: number | null): string {
+    if (score === null) return '—';
+    return String(score);
+  }
   const endonymColumn = { ...EndonymColumn, isInitiallyVisible: true };
   const codeColumn = {
     ...CodeColumn,
@@ -111,6 +149,84 @@ const LanguageTable: React.FC = () => {
           isInitiallyVisible: false,
           sortParam: SortBy.PopulationOfDescendents,
           columnGroup: 'Population',
+        },
+        {
+          key: 'Vitality (Meta)',
+          label: 'Vitality (Meta)',
+          render: (lang) => {
+            const { score, explanation } = computeVitalityMetascore(lang);
+            const b = scoreBucket(score);
+            return (
+              <Hoverable hoverContent={explanation}>
+                <div style={{ display: 'flex', alignItems: 'center', gap: '0.35em', color: bucketColor(b) }}>
+                  {bucketIcon(b)}
+                  <span style={{ fontVariantNumeric: 'tabular-nums' }}>{scoreLabel(score)}</span>
+                </div>
+              </Hoverable>
+            );
+          },
+          isNumeric: true,
+          columnGroup: 'Vitality',
+        },
+        {
+          key: 'Vitality (ISO)',
+          label: 'Vitality (ISO)',
+          render: (lang) => {
+            const all = getAllVitalityScores(lang);
+            const s = all[VitalityMeterType.ISO].score;
+            const b = scoreBucket(s);
+            return (
+              <Hoverable hoverContent={getVitalityExplanation(VitalityMeterType.ISO, lang, s)}>
+                <div style={{ display: 'flex', alignItems: 'center', gap: '0.35em', color: bucketColor(b) }}>
+                  {bucketIcon(b)}
+                  <span style={{ fontVariantNumeric: 'tabular-nums' }}>{scoreLabel(s)}</span>
+                </div>
+              </Hoverable>
+            );
+          },
+          isNumeric: true,
+          isInitiallyVisible: false,
+          columnGroup: 'Vitality',
+        },
+        {
+          key: 'Vitality (Eth 2013)',
+          label: 'Vitality (Eth 2013)',
+          render: (lang) => {
+            const all = getAllVitalityScores(lang);
+            const s = all[VitalityMeterType.Eth2013].score;
+            const b = scoreBucket(s);
+            return (
+              <Hoverable hoverContent={getVitalityExplanation(VitalityMeterType.Eth2013, lang, s)}>
+                <div style={{ display: 'flex', alignItems: 'center', gap: '0.35em', color: bucketColor(b) }}>
+                  {bucketIcon(b)}
+                  <span style={{ fontVariantNumeric: 'tabular-nums' }}>{scoreLabel(s)}</span>
+                </div>
+              </Hoverable>
+            );
+          },
+          isNumeric: true,
+          isInitiallyVisible: false,
+          columnGroup: 'Vitality',
+        },
+        {
+          key: 'Vitality (Eth 2025)',
+          label: 'Vitality (Eth 2025)',
+          render: (lang) => {
+            const all = getAllVitalityScores(lang);
+            const s = all[VitalityMeterType.Eth2025].score;
+            const b = scoreBucket(s);
+            return (
+              <Hoverable hoverContent={getVitalityExplanation(VitalityMeterType.Eth2025, lang, s)}>
+                <div style={{ display: 'flex', alignItems: 'center', gap: '0.35em', color: bucketColor(b) }}>
+                  {bucketIcon(b)}
+                  <span style={{ fontVariantNumeric: 'tabular-nums' }}>{scoreLabel(s)}</span>
+                </div>
+              </Hoverable>
+            );
+          },
+          isNumeric: true,
+          isInitiallyVisible: false,
+          columnGroup: 'Vitality',
         },
         {
           key: 'CLDR Coverage',


src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,55 @@ import PopulationWarning from '../common/PopulationWarning';
...
+    if (score >= 1) return 1; // low
+    return 0; // none/extinct
+  }
+
+  function bucketIcon(bucket: 0 | 1 | 2 | 3) {
+    switch (bucket) {
+      case 3:
+        return <WifiHigh size="1em" />;
+      case 2:
+        return <Wifi size="1em" />;
```
conradarcturus, line 41: Oddly, "WifiHigh" is only 3 bars but "Wifi" is 4 bars. So you should swap them.

Also, the Lucide library offers 2 aliases for the icons, eg. `Wifi` and `WifiIcon`. I recommend using the Icon ones so its clearer what it is.

```suggestion
        return <WifiIcon size="1em" />;
      case 2:
        return <WifiHighIcon size="1em" />;
```

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
...
+                    gap: '0.35em',
+                    color: bucketColor(b),
+                  }}
+                >
+                  {bucketIcon(b)}
+                  <span style={{ fontVariantNumeric: 'tabular-nums' }}>{scoreLabel(score)}</span>
+                </div>
+              </Hoverable>
+            );
+          },
```
conradarcturus, line 183: I realize that this file is getting too big. Can you move the markdown to its own class `LanguageVitalityCell`? It will also help with code reuse.

Thanks for including the hoverable :) I like it.

```suggestion
          render: (lang) => {
            <LanguageVitalityCell lang={lang} type={VitalityMeterType.Metascore} />
            );
          },
```

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
           sortParam: SortBy.PopulationOfDescendents,
           columnGroup: 'Population',
         },
+        {
+          key: 'Vitality (Meta)',
+          label: 'Vitality (Meta)',
```
conradarcturus, line 164: 1) the label is optional if its identical to the key
2) For the Metascore I'd just call this 'Vitality'.

```suggestion
          key: 'Vitality',
```

src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,55 @@ import PopulationWarning from '../common/PopulationWarning';
...
   const { languagesInSelectedSource } = useDataContext();
+
+  // Helper functions for vitality display
+  function scoreBucket(score: number | null): 0 | 1 | 2 | 3 {
+    if (score === null) return 0;
+    if (score >= 7) return 3; // strong
+    if (score >= 4) return 2; // medium
+    if (score >= 1) return 1; // low
+    return 0; // none/extinct
+  }
```
conradarcturus, line 34: Call this `getScoreBucket` instead. Generally its a good idea to make sure all function calls have verbs.

Also, instead of numeric values I recommend using an enum with 5 values strong/medium/low/extinct/unknown.

Also there is way too many lines to include in the LanguageTable class. Generally I try to limit class length. See my recommendation below to move the language vitality cell render to its own class `LanguageVitalityCell`

src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,55 @@ import PopulationWarning from '../common/PopulationWarning';
...
+  function bucketIcon(bucket: 0 | 1 | 2 | 3) {
+    switch (bucket) {
+      case 3:
+        return <WifiHigh size="1em" />;
+      case 2:
+        return <Wifi size="1em" />;
+      case 1:
+        return <WifiLow size="1em" />;
+      default:
+        return <WifiOff size="1em" />;
```
conradarcturus, line 45: Separate extinct & unknown. You will see in your screenshot it looks like all of the dialects & language families are extinct. When there is no vitality score, you should show a grey hyphen instead & make sure the hoverable explains "there is no vitality estimate for this measure for this language".

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
...
+          isNumeric: true,
+          columnGroup: 'Vitality',
+        },
+        {
+          key: 'Vitality (ISO)',
+          label: 'Vitality (ISO)',
+          render: (lang) => {
+            const all = getAllVitalityScores(lang);
+            const s = all[VitalityMeterType.ISO].score;
+            const b = scoreBucket(s);
```
conradarcturus, line 193: Avoid single-letter variable names. I cannot express how much work I had to do in my first job when I inherited a machine-learning pipeline where everything was given names like "a, b, r, r2, rr"...

Also you never use the variable `all` a second time. Sometimes for code readability its good to have intermediate value, but in this case it does not seem necessary.

```suggestion
            const vitalityScore = getAllVitalityScores(lang)[VitalityMeterType.ISO].score;
            const scoreBucket = getScoreBucket(vitalityScore);
```

src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,55 @@ import PopulationWarning from '../common/PopulationWarning';
...
+  // Helper functions for vitality display
+  function scoreBucket(score: number | null): 0 | 1 | 2 | 3 {
+    if (score === null) return 0;
+    if (score >= 7) return 3; // strong
+    if (score >= 4) return 2; // medium
+    if (score >= 1) return 1; // low
+    return 0; // none/extinct
+  }
+
+  function bucketIcon(bucket: 0 | 1 | 2 | 3) {
```
conradarcturus, line 36: Since you are returning react component, you should use a functional component rather than just a function. Also since you reuse the bucket time you should make it an enum.
```suggestion
const BucketIcon: React.FC<{ bucket: VitalityBucket }> = ({ bucket }) => {
```

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
...
+          render: (lang) => {
+            const { score, explanation } = computeVitalityMetascore(lang);
+            const b = scoreBucket(score);
+            return (
+              <Hoverable hoverContent={explanation}>
+                <div
+                  style={{
+                    display: 'flex',
+                    alignItems: 'center',
+                    gap: '0.35em',
```
conradarcturus, line 174: Thanks for using `em` units. However, I recommend using powers of 2 in distances -- it keeps it more standard.
```suggestion
                    gap: '0.25em',
```

src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,55 @@ import PopulationWarning from '../common/PopulationWarning';
...
+        return 'var(--color-text-orange)';
+      default:
+        return 'var(--color-text-red)';
+    }
+  }
+
+  function scoreLabel(score: number | null): string {
+    if (score === null) return '—';
+    return String(score);
+  }
```
conradarcturus, line 65: Instead of making this a function, you can just do `${score ?? '—'}`.

conradarcturus (commented):
Looking really good! I left a bunch of comments, but, really, it's just building on top of what you have and teaching you some good react style paradigms. It's nice to have it started so its easier to visualize.

Btw when you post a test plan I encourage including links so its fast for reviewers to reproduce the tests. Most page params are saved to the URL. Column visibility is not currently saved to the URL though -- if that's not yet a github issue we should file that so its easier to keep it.
http://localhost:5173/lang-nav/data?view=Table&searchBy=English+Name&limit=20&searchString=french

Note in that link (and the screenshot below) I used the text filter "French" to try to get a small list of locales @ various states of vitality so its easier to see side-by-side. It took a bit of experimentation to figure the best filter to quickly get the information. The yellow looks a bit too bright, we should probably decrease the luminosity of --color-text-yellow in the stylesheet `index.css`.

I loaded your PR on my devserver and with the tweaks I recommend it looks like this:

Image

src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,55 @@ import PopulationWarning from '../common/PopulationWarning';
...
+        return <Wifi size="1em" />;
+      case 1:
+        return <WifiLow size="1em" />;
+      default:
+        return <WifiOff size="1em" />;
+    }
+  }
+
+  function bucketColor(bucket: 0 | 1 | 2 | 3): string {
+    switch (bucket) {
```
conradarcturus, line 50: add if null return `var(--color-text-secondary)`

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
...
+                <div
+                  style={{
+                    display: 'flex',
+                    alignItems: 'center',
+                    gap: '0.35em',
+                    color: bucketColor(b),
+                  }}
+                >
+                  {bucketIcon(b)}
+                  <span style={{ fontVariantNumeric: 'tabular-nums' }}>{scoreLabel(s)}</span>
```
conradarcturus, line 205: Instead of showing the score, show the string. You'll need to modify `getVitalityExplanation` to return another value in the `{}`. The Metascore should stay as the number, but the non-meta labels should the original string.

For Ethnologue 2025, we should trim the numbers off the beginning to avoid confusing (the numbers make sense where it came from originally, a google spreadsheet. You can quickly do that with this: `lang.vitalityEth2025?.split(' ')[1]`

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
...
+                </div>
+              </Hoverable>
+            );
+          },
+          isNumeric: true,
```
conradarcturus, line 210: Remove isNumeric -- the big effect of isNumeric is that it right-aligns the columns so its easier to compare values in the thousands. In this case, its easier to read the wifi signals if they are left aligned.

isNumeric is really for multi-digit number-only columns since this:
```
  9
 25
  8
143
```
looks better than this:
```
9
25
8
143
```
Since the ones-column is aligned, the the tens & so for.
```suggestion
```

conradarcturus (changes requested):

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +159,112 @@ const LanguageTable: React.FC = () => {
...
+          columnGroup: 'Vitality',
+        },
+        {
+          key: 'Vitality (ISO)',
+          label: 'Vitality (ISO)',
```
conradarcturus, line 189: probably better to call this Vitality: ISO
```suggestion
          key: 'Vitality: ISO',
```

src/views/language/LanguageTable.tsx
conradarcturus: Also `WifiZeroIcon` is probably better than `WifiOffIcon` -- visually `WifiOffIcon` signifies "not 4" not "0".

src/views/language/LanguageVitalityCell.tsx
```diff
@@ -0,0 +1,99 @@
...
+  if (type === VitalityMeterType.Metascore) {
+    const result = computeVitalityMetascore(lang);
+    vitalityScore = result.score;
+    hover = result.explanation;
+  } else {
+    const all = getAllVitalityScores(lang);
+    vitalityScore = all[type].score;
+    hover = getVitalityExplanation(type, lang, vitalityScore);
+  }
+
```
conradarcturus, line 79: There is a lot here that you already do in other functions. You can compress this all to this:

```suggestion
  const { score: vitalityScore, explanation: hover } = getAllVitalityScores(lang)[type];
```

src/views/language/LanguageVitalityCell.tsx
```diff
@@ -0,0 +1,99 @@
...
+const BucketIcon: React.FC<{ bucket: VitalityBucket }> = ({ bucket }) => {
+  switch (bucket) {
+    case VitalityBucket.Strong:
+      return <WifiIcon size="1em" />;
+    case VitalityBucket.Medium:
+      return <WifiHighIcon size="1em" />;
+    case VitalityBucket.Low:
+      return <WifiLow size="1em" />;
+    case VitalityBucket.Extinct:
+      return <WifiOff size="1em" />;
```
conradarcturus, line 44: Use Icon namespace here too
```suggestion
      return <WifiLowIcon size="1em" />;
    case VitalityBucket.Extinct:
      return <WifiOffIcon size="1em" />;
```

src/views/language/LanguageVitalityComputation.tsx
conradarcturus, line 180: The number works for the metascore, but not the individual scores. As you can see in the screenshot I shared with you, when looking at the individual units it is probably better to show the text label too ( "Institutional", "Living" ...)
```suggestion
export function computeVitalityMetascore(lang: LanguageData): {
score: number | null;
label: string | null;
explanation: React.ReactNode;
} {
```

src/views/language/LanguageVitalityCell.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          gap: '0.25em',
+          color: bucketColor(bucket),
+        }}
+      >
+        <BucketIcon bucket={bucket} />
+        <span style={{ fontVariantNumeric: 'tabular-nums' }}>{vitalityScore ?? '—'}</span>
```
conradarcturus, line 93: use the label here, unless its a metascore, then use the score

src/views/language/LanguageTable.tsx
```diff
@@ -112,6 +117,29 @@ const LanguageTable: React.FC = () => {
           sortParam: SortBy.PopulationOfDescendents,
           columnGroup: 'Population',
         },
+        {
+          key: 'Vitality',
```
conradarcturus, line 121: I apologize, looking at it now its confusing what this means without the subtitle. But maybe something less jargony, like "Overall Vitality"

conradarcturus (changes requested):
Almost there :D

src/views/language/LanguageTable.tsx
abdulkalam556: I'm planning to use Metascore

src/views/language/LanguageTable.tsx
```diff
@@ -14,8 +14,13 @@ import PopulationWarning from '../common/PopulationWarning';
 import { CodeColumn, EndonymColumn, NameColumn } from '../common/table/CommonColumns';
 import ObjectTable from '../common/table/ObjectTable';
 
+import LanguageVitalityCell from './LanguageVitalityCell';
+import { VitalityMeterType } from './LanguageVitalityComputation';
+
 const LanguageTable: React.FC = () => {
   const { languagesInSelectedSource } = useDataContext();
+
+  // (vitality helpers moved into LanguageVitalityCell)
```
conradarcturus, line 23: This comment isn't necessary anymore

src/views/language/LanguageVitalityCell.tsx
```diff
@@ -0,0 +1,86 @@
...
+          alignItems: 'center',
+          gap: '0.25em',
+          color: bucketColor(bucket),
+        }}
+      >
+        <BucketIcon bucket={bucket} />
+        <span>
+          {type === VitalityMeterType.Metascore
+            ? (vitalityScore?.toFixed(1) ?? '—')
+            : (getAllVitalityScores(lang)[type].label ?? '—')}
```
conradarcturus, line 79:
```suggestion
  const { score: vitalityScore, explanation: hover, label } = getAllVitalityScores(lang)[type];
  const bucket = getScoreBucket(vitalityScore);

  return (
    <Hoverable hoverContent={hover}>
      <div
        style={{
          display: 'flex',
          alignItems: 'center',
          gap: '0.25em',
          color: bucketColor(bucket),
        }}
      >
        <BucketIcon bucket={bucket} />
        <span>
          {type === VitalityMeterType.Metascore
            ? (vitalityScore?.toFixed(1) ?? '—')
            : (label ?? '—')}
```

conradarcturus (approved):
2 small edits, other great to merge :)

Issue #188, opened 2025-08-14
Controls: Make language vitality filter


conradarcturus:
People using langnav may want to see specific languages based on their vitality. For example, which are the endangered languages? Or to leave out extinct languages from the data.

## New filters

We want new filters based on vitality. Since we have 3 different vitality scores, we will want 4 selectors, 1 filter function, and integrations in the necessary widgets. The filters for ISO, Ethnologue (coarse), and Ethnologue (fine) should be straightforward -- but the filter for the Metascore is tricky, so we'll ignore it for this task.

If we look at the languages in France, we see languages with lots of different states of vitality. https://translation-commons.github.io/lang-nav/data?view=Table&territoryFilter=france
<img width="850" height="806" alt="Image" src="https://github.com/user-attachments/assets/be3c7f53-8486-4e2d-9336-438ef3817fd8" />

## Filtering in LangNav

Let's look where filters come into play in this screenshot. [link](https://translation-commons.github.io/lang-nav/data?objectType=Locale&view=Table&searchString=per&territoryFilter=Iran)
<img width="1145" height="754" alt="Image" src="https://github.com/user-attachments/assets/07e179e6-6cf2-483e-b3f3-6f934623c06c" />

In the left-hand sidebar there are filters, right now 2 Scope filters and 1 Territory filter. There's also filtering from substrings in the Search Bar.

Filters are used in multiple data views.
* CardList: filters the cards that appear, see where the filters are combined in objectsVisible
* ObjectTable: likewise, filters which rows appear
* Treelist: works a bit differently -- all nodes are hidden except ones with descendents that pass the filter. Nodes that pass the filter continue showing all descendents. You'll make your update in `TreeListPageBody.tsx`. Scope is handled separately

Filters then are used in multiple widgets
* pathNav/ObjectPath shows the active filters (see `FilterPath.tsx`)

## Page parameters

The first thing you need to do, is to add a way to store the filter states. We'll use the PageParam concepts -- then the filters are stored in the page URL. Add 3 new parameters to `PageParamsTypes.tsx` -- similar to the scope parameters. Check where `languageScopes` is specified to see where you will to specific the vitality levels. I just converted the vitality levels to enums so you should able to use them easily.

We will want 3 multi-selector UI elements, use `LanguageScopeSelector` as your template.

[...]

PR #264, opened 2025-10-27
Feat/vitality filter 188

abdulkalam556:
## 🧪 Overview  
This PR implements a new *“Vitality” filter* for the navigation UI, as requested in issue #188.  
We’ve added a selector component for vitality, integrated it into the side panel, passed it through page-parameter logic, and updated all relevant views so that the navigation/listing respects the vitality filter.

---

## ✅ What’s been done  
- Introduced a `VitalitySelector` component to allow users to choose vitality levels.  
- Added UI elements to the side-panel so the vitality selector appears alongside existing filters.  
- Extended the page/URL parameters logic so the selected vitality value is captured in the navigation query.  
- Updated the filter path component so vitality is passed into the backend/API or selector logic.  
- Updated all relevant views so the vitality filter is applied correctly in listings.  
- Added tests and lint adjustments to cover the new functionality.  

---

### 🔗 URL to test  
[http://localhost:5173/lang-nav/data?view=Table&territoryFilter=france&vitalityISO=9%2C1%2C0&vitalityEth2013=9%2C2%2C4%2C5](http://localhost:5173/lang-nav/data?view=Table&territoryFilter=france&vitalityISO=9%2C1%2C0&vitalityEth2013=9%2C2%2C4%2C5)

### 🖼️ Screenshot after changes  
Screenshot 2025-10-26 at 11 45 31 PM

### DIFF
--- a/src/tests/MockPageParams.test.tsx
+++ b/src/tests/MockPageParams.test.tsx
@@ -33,6 +33,9 @@ export const createMockUsePageParams = (
   languageSource: LanguageSource.All,
   localeSeparator: LocaleSeparator.Underscore,
   territoryFilter: '',
+  vitalityISO: [],
+  vitalityEth2013: [],
+  vitalityEth2025: [],
   ...overrides,
 });
 


src/features/page-params/getParamsFromURL.ts
```diff
@@ -35,6 +54,16 @@ export function getParamsFromURL(urlParams: URLSearchParams): PageParamsOptional
         if (value === '[]') params[key] = [];
         else params[key] = value.split(',').filter(Boolean) as TerritoryScope[];
         break;
+      // Add new vitality array parsing
```
conradarcturus: Don't include developmental comments like "Add new vitality array parsing". You want to write evergreen code, not code that becomes stale.

```suggestion
```

PR #299, opened 2025-11-10
Improve exotic characters in search 242

Maheedhar18:
## Summary

* Implemented accent-agnostic search to improve search functionality and highlighting.
* Searching for alphanumeric IDs now works for English Name or ID Search mode.

## Changes

`stringUtils.ts` - 

* Added normalizeAccents() helper function that removes accent marks.
* Updated anyWordStartsWith() to normalize both the search query and text being searched, making matching accent-insensitive and case-insensitive.
* Changed the word-splitting regex to split on characters that are neither letters nor numbers. This keeps alphanumeric codes like "jara1263" together as single words, making them searchable in the NameOrCode format.

`Highlightable.tsx` -
* Modified to normalize text and search patterns before regex matching then map the match positions back to the original un-normalized text for display.
* This allows searches like "Cote" to match and highlight "Côte" while preserving the accented characters in the output.

`filter.tsx`, `ObjectField.tsx` -

* Simplified query handling since normalization is now handled in anyWordStartsWith()
* Removed redundant .toLowerCase() calls

## Screenshots
Screenshot 2025-11-10 at 1 15 07 PM
Screenshot 2025-11-10 at 1 14 52 PM
Screenshot 2025-11-10 at 1 13 48 PM
Screenshot 2025-11-10 at 1 14 39 PM
Screenshot 2025-11-10 at 1 14 07 PM


src/shared/lib/stringUtils.ts
```diff
@@ -30,15 +30,28 @@ export function joinOxfordComma(strs: string[]): string {
...
  * Sometimes inputted queries may contain whitespace -- for that cause we check if the general
  * input also matches.
+ * Both the string and query are normalized to be accent-insensitive and case-insensitive.
  */
-export function anyWordStartsWith(str: string, lowercaseQuery: string) {
+export function anyWordStartsWith(str: string, query: string) {
   const strLowercase = str.toLowerCase();
+  const queryLowercase = query.toLowerCase();
+  const normalizedStr = normalizeAccents(strLowercase);
+  const normalizedQuery = normalizeAccents(queryLowercase);
```
conradarcturus: Sometimes its good to split variable declarations for clarity. Here, you can combine them.

```suggestion
  const normalizedStr = normalizeAccents(str.toLowerCase());
  const normalizedQuery = normalizeAccents(query.toLowerCase());
```

Issue #317, opened 2025-11-21
Maps: Add scaleBy page parameter to scale map nodes


conradarcturus:
The circles on the map are all the same size. However, you may want to scale them by the land area of a territory or the population. Fortunately, `SortBy` and its related functions already give us a huge starting point since it converts objects to sortable values (usually, but not always, numbers). The alphabetic SortBy values can be used in coloring, but really wouldn't make any sense for sizing. However, it's a lot easier

* Make a new "scales" subfolder in the `@features/transforms` folder
* Add a `ScaleBy` enum, can follow the same as `ColorBy` -- where it extends `SortBy` but adds a new value (None).
* Add a page parameter (key, optional param, instanced param) in `PageParamTypes.tsx`
  * There will be many switch statements to update before the param works. It's easy to miss updating it in `getParamsFromURL`
* Add a dropdown selector for this param, just like `SortBySelector` and `ColorBySelector`
  * put the selector in the `SidePanel` right after ColorBy.
* Make a `useScale` function that will be similar to `useColor` but not quite as complex. For now, I don't think we need a colorBar equivalent so you don't need to pass as many return values -- definitely `getScale`.
* Use the function in `MapCircles` to modify the scale of the circles. You'll probably want to take the square root of the value with `x ** 0.5`.

Try scaling circles by population (languages, territories), land area (for territories specifically) and see how it looks.

Recommendation: See how `colorBy` is used and replicate most of that functionality. Fortunately we don't need to map the scale to a gradient.

PR #325, opened 2025-11-30
317 feature/add scaling for map nodes

cchandram:
### Summary ###

Issue #317 

This pull request introduces a new "Scale By" feature that allows map circles to be dynamically scaled based on selected data fields (such as population or count of languages). The implementation includes UI updates, parameter handling, and a new scaling logic hook, as well as tests to ensure correct scaling behavior. 

Map circles can be scaled by population, count of languages (when mapping territories), and population attested as well as count of territories (when mapping languages).

**Changes:**

**1. Scale By Parameter and UI Integration**
- Added a new `scaleBy` parameter to `PageParams`, with corresponding updates to types, defaults, and URL handling. This allows users to select a field by which map circles are scaled.
- Introduced a new `ScaleBySelector` component in the side panel, letting users choose the scaling field from allowed options.

**2. Scaling Logic Implementation**
- Added `useScale` hook to compute the appropriate radius multiplier for each object based on the selected `scaleBy` field, with support for logarithmic and square-root scaling for certain fields.
- Updated `MapCircles` and `HoverableCircle` components to use the scaling logic, so circles are drawn with the correct radii according to the selected scale. 

### Testing ###
Added unit tests for the scaling logic to verify correct behaviour for different fields and edge cases:
- default multiplier when scaleBy is undefined or 'None'
- latitude mapping (-90..90) -> radius multiplier 2..10
- population mapping uses sqrt + log transform and handles missing values (non-renderable)

### Screenshots ###
lang-nav-map-circles-scaled-1
lang-nav-map-circles-scaled-2

"Scale By" dropdown when mapping territories:
lang-nav-map-circles-scaled-4

"Scale By" dropdown when mapping languages:
lang-nav-map-circles-scaled-6

### DIFF
--- a/src/features/params/PageParamTypes.tsx
+++ b/src/features/params/PageParamTypes.tsx
@@ -1,6 +1,6 @@
 import { ColorBy, ColorGradient } from '@features/transforms/coloring/ColorTypes';
-import { SortBehavior, SortBy } from '@features/transforms/sorting/SortTypes';
 import { ScaleBy } from '@features/transforms/scales/ScaleTypes';
+import { SortBehavior, SortBy } from '@features/transforms/sorting/SortTypes';
 
 import { LanguageScope, LanguageSource } from '@entities/language/LanguageTypes';
 import {
--- a/src/widgets/controls/SidePanel.tsx
+++ b/src/widgets/controls/SidePanel.tsx
@@ -10,7 +10,6 @@ import {
 } from '@features/params/ui/SelectorDisplayContext';
 import ColorBySelector from '@features/transforms/coloring/ColorBySelector';
 import ColorGradientSelector from '@features/transforms/coloring/ColorGradientSelector';
-import ScaleBySelector from '@features/transforms/scales/ScaleBySelector';
 import LanguageFilterSelector from '@features/transforms/filtering/LanguageFilterSelector';
 import LanguageScopeSelector from '@features/transforms/filtering/LanguageScopeSelector';
 import TerritoryFilterSelector from '@features/transforms/filtering/TerritoryFilterSelector';
@@ -21,6 +20,7 @@ import {
   VitalityEth2025Selector,
 } from '@features/transforms/filtering/VitalitySelector';
 import WritingSystemFilterSelector from '@features/transforms/filtering/WritingSystemFilterSelector';
+import ScaleBySelector from '@features/transforms/scales/ScaleBySelector';
 import SortBySelector from '@features/transforms/sorting/SortBySelector';
 import SortDirectionSelector from '@features/transforms/sorting/SortDirectionSelector';
 


src/features/transforms/scales/__tests__/useScale.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        const latMax = getBaseLanguageData('max', 'LatMax');
+        latMax.latitude = 90;
+
+        const latMin = getBaseLanguageData('min', 'LatMin');
+        latMin.latitude = -90;
+
+        const { result } = renderHook(() => useScale({ objects: [latMax, latMin], scaleBy: SortBy.Latitude }));
+
+        // Max latitude should map to the maximum multiplier (10)
+        expect(result.current.getScale(latMax)).toBeCloseTo(10);
```
conradarcturus: oh! I didn't know about `toBeCloseTo` -- that's good because of float errors sometimes :p

src/features/transforms/scales/useScale.tsx
```diff
@@ -0,0 +1,168 @@
...
+            if (numericValue <= 0) return 0;
+            if (shouldUseLogScale) return Math.log10(numericValue) / range;
+            return numericValue / range;
+        },
+        [scaleBy, tMin, tMax, range, shouldUseLogScale],
+    );
+
+    const getScale = useCallback(
+        (object: ObjectData) => {
+            if (!scaleBy || scaleBy === 'None') return 2; // default radius multiplier
```
conradarcturus: leave this as 2

src/features/transforms/scales/useScale.tsx
```diff
@@ -0,0 +1,168 @@
...
+        },
+        [scaleBy, getNormalizedValue],
+    );
+
+    return { scaleBy, getScale, maxValue, minValue };
+};
+
+export default useScale;
+
+function getMinimumValue(scaleBy?: ScaleBy): number {
```
conradarcturus: I'd move the functions you copied from `useColor` and just put it in the `@features/transforms` directory (the parent one to this) make a separate `getMinimumValue.ts` and `getMaximumValue.ts` (or maybe put them in the same file?)

But how do you handle both a ScaleBy and a ColorBy? very optimistically think you can do the suggestions below.
```suggestion
export function getMinimumValue(field?: ColorBy | ScaleBy): number {
```

src/features/transforms/coloring/ColorTypes.ts
```diff
@@ -1,7 +1,57 @@
...
+  ];
+
+  let filteredColorBys = [...allowedColorBys];
+
+  // Exclude area and count-of-languages when mapping Languages (use CountOfDialects instead)
+  if (objectType === ObjectType.Language) {
+    filteredColorBys = filteredColorBys.filter(
+      (s) => s !== SortBy.CountOfLanguages && s !== SortBy.Area,
+    );
+  }
```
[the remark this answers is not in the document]
conradarcturus: I made a task to follow up on this -- let me know if you are interested in it or I can just leave it unassigned. https://github.com/Translation-Commons/lang-nav/issues/333

PR #340, opened 2025-12-25
Refactor: Consolidate logic for get*BysApplicableToObjectType 

### LINKED ISSUE #333: Consolidate logic for get*BysApplicableToObjectType
We have `getSortBysApplicableToObjectType`, `getColorBysApplicableToObjectType` and soon `getScaleBysApplicableToObjectType`.

Right now, there are many manual updates, which leaves it prone for error. If we modify an object type (like by adding a new field) we have to manually update each of these. If we change how things are visualized, etc.

Rather, it may be better to separate it into 2 steps that are intersected:
* getFieldsForObjectType
* getFieldsForColoring, getFieldsForSorting, getFieldsForScaling

cchandram:
This pull request refactors how field applicability (for sorting, colouring, and scaling) is determined across object types. The logic is now centralised in a new utility module, removing duplicated and scattered logic from multiple files. This makes the codebase easier to maintain and extend.

The refactor organises the previous _get*BysApplicableToObjectType_ into two steps: `intersectAllowedWithObjectType()` and `getFieldsFor*()`

For example, in the case of colouring we now call the functions like this:
`const applicableColorBys = intersectAllowedWithObjectType(getFieldsForColoring(), objectType);`

**Changes**
* Added a new `FieldApplicability.ts` module that defines which fields are available for sorting, colouring, and scaling for each object type, and provides utility functions to compute the intersection of allowed fields with those available for a given object type.
* Removed legacy field applicability logic from `getSortBysApplicableToObjectType.ts`, `ColorTypes.ts`, and `ScaleTypes.ts`, consolidating it into the new module.
* Updated `SortBySelector.tsx`, `ColorBySelector.tsx`, `ScaleBySelector.tsx`, and `TreeListOptions.tsx` to use the new centralised utility functions for determining applicable fields, replacing previous ad-hoc or duplicated logic. 

This is only a starting point and could probably be improved upon. Not sure if nesting getFieldsFor*() is the most elegant solution, but it allowed for minimal code duplication.

Closes #333.

### DIFF
--- a/src/features/treelist/TreeListOptions.tsx
+++ b/src/features/treelist/TreeListOptions.tsx
@@ -110,7 +110,7 @@ const ShowDataSelector: React.FC<{
   setShowData: (value: ColorBy) => void;
 }> = ({ showData, setShowData }) => {
   const { objectType } = usePageParams();
-  const applicableSortBys: ColorBy[] = ['None', ...getSortBysApplicableToObjectType(objectType)];
+  const applicableFields: ColorBy[] = ['None', ...getSortBysApplicableToObjectType(objectType)];
 
   return (
     <Selector<ColorBy>


src/features/transforms/fields/FieldApplicability.ts
```diff
@@ -0,0 +1,147 @@
...
+ */
+export function getFieldsForObjectType(objectType: ObjectType): SortBy[] {
+  const specific = getSpecificFieldsForObjectType(objectType);
+  const deduped = new Set<SortBy>([...COMMON_FIELDS, ...specific]);
+  return [...deduped];
+}
+
+/** Fields that make sense as sorting inputs */
+export function getFieldsForSorting(): SortBy[] {
+  // Currently all SortBy fields are sortable; can be restricted in future
```
conradarcturus: sgtm

src/features/transforms/scales/ScaleBySelector.tsx
```diff
@@ -3,14 +3,18 @@ import React from 'react';
...
 
-import { getScaleBysApplicableToObjectType, ScaleBy } from './ScaleTypes';
+import { ScaleBy } from './ScaleTypes';
 
 const ScaleBySelector: React.FC = () => {
   const { scaleBy, updatePageParams, objectType } = usePageParams();
   const { display } = useSelectorDisplay();
 
-  const applicableScaleBys = getScaleBysApplicableToObjectType(objectType);
+  const applicableScaleBys = intersectAllowedWithObjectType(getFieldsForScaling(), objectType);
```
conradarcturus: (here and elsewhere) It's a bit funny to import 2 functions and require the user to call `intersectAllowedWithObjectType`. I'd probably make `getFieldsForScaling` and  `intersectAllowedWithObjectType` private in FieldApplicability and have the publicly exported function `getScaleBysApplicableToObjectType`

PR #343, opened 2025-12-29
Refactor: Moved largest descendant computation to data loading (new)

### LINKED ISSUE #241: Refactor: Most "largest descendent language" algorithm to data loading step
Right now the algorithm is run for the `LanguagesLargestDescendant` report. However we're mutating data when we should be view it.

Rather, we should run it when we run the other `DataAssociations` in `CoreData` -- like around where we call `computeOtherPopulationStatistics`

cchandram:
This pull request refactors how the "largest descendant language" is computed and stored, moving the computation out of the UI layer and into the data processing pipeline. This improves performance and code organization by ensuring these calculations are done once during data loading rather than on every component render.

(New version of now deleted PR #342 after #337 introduced changes to `CoreData.tsx`, which I found easier to deal with in a new branch instead of handling merge conflicts and having a slightly messy PR)

### Changes: Data processing improvement

   * Moved the algorithm into `src/features/data/compute/computeLargestDescendant.ts`
   * Integrated it in `connectObjects.ts` 
   * Removed the redundant computation from the LanguagesLargestDescendant report component

Closes #241.

### DIFF
--- a/src/features/data/compute/computeLargestDescendant.ts
+++ b/src/features/data/compute/computeLargestDescendant.ts
@@ -1,0 +1,41 @@
+import { LanguageData } from '@entities/language/LanguageTypes';
+
+export function computeLargestDescendant(languages: LanguageData[]): void {
+  // Clear the largest descendants first since it may change a lot if the schema changes.
+  languages.forEach((lang) => {
+    lang.largestDescendant = undefined;
+  });
+
+  // Compute the largest descendants for each language
+  languages.forEach((lang) => {
+    lang.largestDescendant = getLargestDescendant(lang);
+  });
+}
+
+function getLargestDescendant(language: LanguageData): LanguageData | undefined {
+  // If it has already been computed, return it.
+  if (language.largestDescendant !== undefined) {
+    return language.largestDescendant;
+  }
+  if (language.childLanguages.length === 0) {
+    return undefined;
+  }
+
+  // We are using populationRough because we want to only use descendants that have a directly
+  // sourced population (aka no language families). Dialects rarely but sometimes have directly
+  // sourced populations.
+  return language.childLanguages.reduce<LanguageData | undefined>((largest, child) => {
+    const childsLargest = getLargestDescendant(child);
+    const candidateLargest =
+      (childsLargest?.populationRough || 0) > (child.populationRough || 0) ? childsLargest : child;
+    if (
+      candidateLargest != null &&
+      candidateLargest.populationRough != null &&
+      candidateLargest.populationRough > 0 &&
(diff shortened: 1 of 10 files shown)


src/features/data/compute/connectObjects.ts
```diff
@@ -36,4 +37,5 @@ export function connectObjectsAndCreateDerivedData(
   createRegionalLocales(territories, locales); // create them after connecting them
   searchLocalesForMissingLinks(locales); // try to find missing links after creating new locales
   computeDescendantPopulation(languagesBySource, writingSystems);
+  computeLargestDescendant(languagesBySource);
```
conradarcturus: While it makes sense to do here -- we should actually add this code to `updateObjectCodesNameAndPopulation.ts` (as that function has evolved we can also rename it to `updateObjectsBasedOnDataParams` since it also update language connections, vitality). For example, Glottolog, ISO, and CLDR have slightly different language families.

As you have it now, `connectObjects` happens only once, on load, when the default language source is `Combined`. But when the language source is changed, `computeLargestDescendant` should also change, particularly since `updateParentsAndDescendants` happens in `updateObject...`. ALSO `updatePopulations` may change language estimates (also, see my comment on populationRough).

src/features/data/compute/computeLargestDescendant.ts
```diff
@@ -0,0 +1,44 @@
...
+    return undefined;
+  }
+
+  // We are using populationRough because we want to only use descendants that have a directly
+  // sourced population (aka no language families). Dialects rarely but sometimes have directly
+  // sourced populations.
+  return children.reduce<LanguageData | undefined>((largest, child) => {
+    const childsLargest = getLargestDescendant(child);
+    const candidateLargest =
+      (childsLargest?.populationRough || 0) > (child.populationRough || 0) ? childsLargest : child;
```
conradarcturus: The world has changed and my original comment is running on old data. If its not a bother I'd love for you to redo the algorithm here.

Instead of using `populationRough` -- the `populationEstimate` field is way better now. Not perfect, but it will have a better baseline for the population of these languages (noting the comment about only doing this after we've updated language populations). populationEstimate takes in the latest census data amongst other things.

Furthermore, instead of implicitly saying (no language families) -- let's just directly rule them out, if scope == Family, skip.

I doubt it will change that many results, but it could.

conradarcturus (changes requested):
Looks great and it definitely does the job -- but since the task was initially published and now there have been some improvements to the underlying data.

conradarcturus (approved):
I force rebased to fix the merging issue -- so if you edit this you may have to delete it from your local server and download it from the repo to avoid terrifying merge conflicts. There's style nit I'd recommend but otherwise lgtm!

Issue #98, opened 2025-07-01
Data quality: Review dubious languages


conradarcturus:
In https://github.com/Translation-Commons/lang-nav/pull/73 we added a report on dubious languages. Let's act on them now

See the report here: https://translation-commons.github.io/lang-nav/?view=Reports

![Image](https://github.com/user-attachments/assets/eb983554-ec37-425c-ad7b-d1e81d7dd485)

There already are recommendations in the report. Ask Conrad for more help if its not clear how to make the fix. It may be good to schedule a 1:1.

For the more part, we'll want to remove lines from languages.tsv
https://github.com/Translation-Commons/lang-nav/blob/master/public/data/languages.tsv#L8106-L8120

In some cases they should be stricken completely. Nushu let's leave for now (since there's no alternative). If you change an entry from a language to a locale, see if there's any information to transfer (eg. a population estimate) and add it as a locale row. We may need to debug it a bit (can we add a locale without a territory? Probably but maybe want to add the territory.
https://github.com/Translation-Commons/lang-nav/blob/master/public/data/locales.tsv

PR #354, opened 2026-01-04
Data Quality: removing some dubious languages

cchandram:
## Summary

Partially covers #98 

This pull request primarily removes some of the languages from `languages.tsv` that are found in the dubious languages report. Additionally, it adds a new Locale found in the report.

### Data changes:

* Removed redundant Locales / orthographic variants / dialects from `public/data/languages.tsv`, including English (e.g., `xxx-en-US`, `xxx-en-GB`), German, French, Portuguese, Cornish, Ladin, Lillooet, and others.
* Added "tuo_CO Tucano (Colombia)" to `public/data/locales.tsv`

Each commit concerns one language. "Redundant" means they are already part of our data and can be found in LangNav. Some of these required a minor amount of research, like checking the Wikipedia entry for Lillooet.

After this PR the dubious languages report will still cover 66 languages, of which most seem to be alternative writing systems. However, since I wasn't confident enough concerning some languages, there are likely a few more redundant ones. For example, I suspect `xxx-fr-x-nor` is the Langue d'oïl _Normand_.

### DIFF
--- a/public/data/locales.tsv
+++ b/public/data/locales.tsv
@@ -4847,7 +4847,6 @@ and_ID	Ansus (Indonesia)	bahasa Ansus	Ethnologue	4,600
 aot_IN	Atong (India) (India)		Ethnologue	4,600	
 mxn_ID	Moi (Indonesia)	bahasa Moi	Ethnologue	4,600	
 tuo_BR	Tucano (Brazil)	Dahseyé	Ethnologue	4,600	recognized_regional
-tuo_CO	Tucano (Colombia) Dahseyé
 ybe_CN	Yellow Uyghur (China)		Ethnologue	4,600	
 nkx_NG	Nkoroo (Nigeria)		Ethnologue	4,550	
 tlv_ID	Taliabu (Indonesia)	bahasa Taliabo	Ethnologue	4,520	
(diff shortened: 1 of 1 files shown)


public/data/languages.tsv
```diff
@@ -8163,47 +8135,23 @@ xxx-jv-x-bms		Banyumasan							0
...
-xxx-pt-CV		Portuguese (Cape Verde)							0					
-xxx-pt-GW		Portuguese (Guinea Bissau)							0					
-xxx-pt-MZ		Portuguese (Mozambique)							0					
-xxx-pt-PT		Portuguese (Portugal)							0					
-xxx-pt-ST		Portuguese (São Tomé and Príncipe)							0					
-xxx-pt-TL		Portuguese (Timor-Leste)							0					
 xxx-ru-Latn		Russian (Latin)							0					
-xxx-scn-x-tara		Tarantino							0					
-xxx-sco-ulster		Ulster Scots							0					
-xxx-sco-x-scotland		Scots (Scotland)							0
```
conradarcturus: Hmmm, while it looks funny perhaps this is referring to the variant tag `scotland` which we only have a locale `eng_scotland` for right now.

PR #361, opened 2026-01-12
Refactor URL param handling and clear context-dependent params

### LINKED ISSUE #321: Page params: Clear some parameters when switching context
## Problem

We have global parameters in langnav. While that keeps a great level of consistency, sometimes it has drawbacks when you switch views and you didn't intend to keep a view-specific parameter.

### Example: Changing views, the limit parameter

Different views have different recommended limits. Viewing a table if you see thousands of rows, the performance is impacted so we recommend decreasing the page item limit.

Image

However, there may also be the case the limit is too low. In particular, when people start at LangNav and search for a language, we intentionally open the Card view with a limit of 1. The idea is that people could use the page controls to flip through other results. But if you switch from that to the table view -- you continue to only see 1 result.

|Step|Screenshot|
|--|--|
|Open [LangNav intro](https://translation-commons.github.io/lang-nav/intro) and type in arabic|Image
|Clicked GO, viewing the card list view with 1 result shown at a time [link](https://translation-commons.github.io/lang-nav/data?searchString=Arabic&limit=1)|Image
|Click to see the next item in the pagination selector [link](https://translation-commons.github.io/lang-nav/data?searchString=Arabic&limit=1&page=2)|Image
|Switch to table view, awkward, only 1 result is showing. You would expect the limit to go back to the default. [link](https://translation-commons.github.io/lang-nav/data?view=Table&limit=1&searchString=arabic)|Image

### Other contextual parameters

Play around with the LangNav and see if there are any other global parameters that don't make sense to keep when changing views, for example the `page` and `colorBy`. Most though make sense to keep, like a territory filter or searchString (eg. we still want languages with the substring "Arabic" in Egypt). However, when changing the `objectType` we may want to change the `page` but the `limit` can change since we're still looking at the same view.

|Changing Parameter|Parameters to clear|Parameters to let be|Special
|--|--|--|--|
|`view`|❌`limit`, ❌`page`, ❌`colorBy`|✅`searchString`, ✅`territoryFilter`, ✅`objectType`|
|`objectType`| ❌`page`|✅`limit`, ✅`territoryFilter`, ✅`colorBy`|�`view`, �`searchString`**

Play around with LangNav and try different combinations that you think should switch. Note: if a user ever makes a mistake they can always press the browser back button and it will revert the change.

** I'm not sure about `searchString` though -- eg. if you switch from Language to Locale or Writing System "Arabic" is still useful -- but not when switching to a Territory. I think we'll handle that in a different PR where we change the `searchString` to the appropriate `languageFilter` etc.

## Recommendation

Most of your changes will be in `getNewURLSearchParams.tsx`.

1. It's too long of a function to begin with, can you refactor the function to have 2 child functions `buildNextURLSearchParams` (lines 14 to 55) and `clearDefaultParams` (lines 57 to 81)? These functions can stay in the file because we don't plan to export it.
2. Add a new function `clearContextDependentParams`
   1. When `newParams.view` is not `undefined`... (and probably when `newParams.view !== prev.get('view')`)
      1. If `newParams.limit` is `undefined` delete the previous limit, you don't need to check it, you can just run `next.delete('limit')` -- it will now revert to the default
      2. Also clear the `page`, `colorBy` (if they are not in `newParams`)
      3. Clear other params you think should be
    2. When `newParams.objectType` is changing...
      1. Clear `page` (if not in `newParams`)
      2. If the previous `view` is View.Details and `newParams.view` is not defined, Then we should switch to the default view (delete the `view` param, the default isn't always the card list)
3. Test it out, try changing different page params. If you like the results post a PR with a test plan :D

OHMFOHS:
Refactored getNewURLSearchParams to automatically clear context-dependent parameters when users switch views or object types, preventing inconsistent URL states and improving user experience.

## Summary

- [x] Mention the issue: fixes #321 

## Testing

- [x] `npm run lint`
- [x] `npm run test`
  - [] Tests added or updated for changed logic
- [x] `npm run build`
- [] Write comments on manual testing how you tested in this PR
- [x] Include screenshots in the changes section below

### Visual changes

- [x] Add before and after screenshots in the table below.
  - [x] Drag and drop images in the GitHub PR comment box to upload screenshots
- [x] Purely new views can just include the "after" screenshot.
- [x] Since more views can be reproduced by just sharing the URL -- add links (eg. [link](https://translation-commons.github.io/lang-nav/data)) to the relevant page and/or conditions to reproduce the view.

| Page/View | Changes | Before | After |
| :--- | :--- | :--- | :--- |
| [Language Card view](https://translation-commons.github.io/lang-nav/data?view=Cards&limit=1) | Clean `limit=1` after switching views |  |  |
| [Map View](https://translation-commons.github.io/lang-nav/data?searchString=arabic&limit=1&view=Map) | Limit updates correctly in the new branch |  |  |

### Data changes

- [ ] TSV edits in `public/data/`
- [ ] Load/connect/compute updates in `src/features/data/`

### Internal changes

- [ ] Logical changes
- [ ] Refactors, moving files around
- [ ] If you notice any changes that require explanations, make sure to include the explanations in the code as well.

## Docs

- [n/a] Updated markdown readme files documenting how the code behaves or how to develop in case there are any relevant changes to make.

### DIFF
--- a/src/features/params/getNewURLSearchParams.ts
+++ b/src/features/params/getNewURLSearchParams.ts
@@ -4,6 +4,10 @@ import { ColorBy } from '@features/transforms/coloring/ColorTypes';
 import { ObjectType, PageParamKey, PageParamsOptional, View } from './PageParamTypes';
 import { getDefaultParams, ProfileType } from './Profiles';
 
+/**
+ * This is used to make the next version of the page parameters in the URL.
+ */
+
 function buildNextURLSearchParams(
   newParams: PageParamsOptional,
   next: URLSearchParams,
@@ -87,29 +91,21 @@ function clearContextDependentParams(
   next: URLSearchParams,
   prev?: URLSearchParams,
 ): URLSearchParams {
-  const prevView = prev?.get('view');
-  const prevObjectType = prev?.get('objectType');
+  const prevOrDefault = getDefaultParams(
+    prev?.get('objectType') as ObjectType,
+    prev?.get('view') as View,
+    prev?.get('profile') as ProfileType,
+    prev?.get('colorBy') as ColorBy,
+  );
 
-  if (newParams.view !== undefined && newParams.view !== prevView) {
-    if (newParams.limit === undefined) {
-      next.delete('limit');
-    }
-    if (newParams.page === undefined) {
-      next.delete('page');
-    }
-    if (newParams.colorBy === undefined) {
-      next.delete('colorBy');
-    }
+  if (newParams.view !== undefined && newParams.view !== prevOrDefault.view) {
+    if (newParams.limit == null) next.delete('limit');
+    if (newParams.page == null) next.delete('page');
+    if (newParams.colorBy == null) next.delete('colorBy');
   }
 
-  if (newParams.objectType !== undefined && newParams.objectType !== prevObjectType) {
-    if (newParams.page === undefined) {
-      next.delete('page');
-    }
-  }
-
-  if (prevView === View.Details && newParams.view === undefined) {
-    next.delete('view');
+  if (newParams.objectType !== undefined && newParams.objectType !== prevOrDefault.objectType) {
+    if (newParams.page == null) next.delete('page');
   }
 
   return next;


src/features/params/getNewURLSearchParams.ts
```diff
@@ -81,3 +81,47 @@ export function getNewURLSearchParams(
 
   return next;
 }
+
+function clearContextDependentParams(
+  newParams: PageParamsOptional,
+  next: URLSearchParams,
+  prev?: URLSearchParams,
+): URLSearchParams {
+  const prevView = prev?.get('view');
```
conradarcturus: one interesting bug I see is that if the previous view is unspecified (so its the default CardList) and a new update sets it directly to CardList then it passes the if statement on like 93 -- however we haven't actually changed the view (it was just default before and will be default after). So the best way is to uplift the previous params to fill in the nulls.

You can use getDefaultParams to get the default states
```suggestion
  const prevOrDefault = getDefaultParams(
    prev?.get('objectType') as ObjectType,
    prev?.get('view') as View,
    prev?.get('profile') as ProfileType,
    prev?.get('colorBy') as ColorBy,
  );
```

src/features/params/getNewURLSearchParams.ts
```diff
@@ -81,3 +81,47 @@ export function getNewURLSearchParams(
...
+  if (newParams.view !== undefined && newParams.view !== prevView) {
+    if (newParams.limit === undefined) {
+      next.delete('limit');
+    }
+    if (newParams.page === undefined) {
+      next.delete('page');
+    }
+    if (newParams.colorBy === undefined) {
+      next.delete('colorBy');
+    }
```
conradarcturus: 2 style nits: 1) when if statements are very small and code is readable, then use the compact 1-line syntax 2) [potentially controversial] I never use strict equality for null or undefined (`=== null` or `=== undefined`), I use approximate equality to null (`== null`) because it handles both cases.

PR #430, opened 2026-02-13
Add icons to WritingSystemCard fields 

### LINKED ISSUE #406: CardList: Change WritingSystemCard to use Icons
Just like #398 , you'll update [WritingSystemCard](https://github.com/Translation-Commons/lang-nav/blob/master/src/entities/writingsystem/WritingSystemCard.tsx) to use Lucide Icons instead of `<label>`s as the labels of sections.

|LanguageCard (the style goal we want)|Writing System card (Current, before changing it)|
|--|--|
|Image|Image

Unique for WritingSystemCard: 
* Order the fields: Sample, Scope, Population, TextDirection, UnicodeSupport. Unlike what we do now, include rows even if they don't have data -- just include a null state with an element like Deemphasized
* For Population (Upper Bound) just keep the title as "Population" and provide a description: "This is a very rough estimate based on adding the populations that use the languages for this writing system, it's likely 50% to 400% off of the actual estimate."
  * If population < 100 say `<Deemphasized>Unknown</Deemphasized>` -- estimates that low are dubious at best.
* The Subtitle will eventually change from an alternative English name to the Endonym, but don't worry about changing that in this particular task since updates to that will be impacted by #400 
* Have TextDirection replace the existing right to left distinction. Show "Right to left" or "<Deemphasized>Left to right</Deemphased>` since RTL writing systems are important to highlight
* Combine the 2 scope cases into 1 scope CardField that adds the extra details for the Variant & Group scopes. For now, the icon BlocksIcon seems to be a good one to capture the various Scope definitions.

Tips:
* Look for icons you think make sense in [the Lucide library](https://lucide.dev/icons/)
* Make informative descriptions, if you have questions we can iterate on the PR.
* Follow the changes that were done in #398
* Doubly make sure to include screenshots in your pull request so we can discuss the icons.

mchen11077:
Fixes #406 

截屏2026-02-12 下午8 49 28

### DIFF
--- a/src/entities/writingsystem/WritingSystemCard.tsx
+++ b/src/entities/writingsystem/WritingSystemCard.tsx
@@ -1,17 +1,24 @@
+import {
+  BlocksIcon,
+  CodeIcon,
+  LanguagesIcon,
+  PencilLineIcon,
+  PilcrowLeftIcon,
+  UsersIcon,
+} from 'lucide-react';
 import React from 'react';
 
-import PopulationWarning from '@widgets/PopulationWarning';
-
+import Hoverable from '@features/layers/hovercard/Hoverable';
 import HoverableObjectName from '@features/layers/hovercard/HoverableObjectName';
 import usePageParams from '@features/params/usePageParams';
-import { getScopeFilter } from '@features/transforms/filtering/filter';
 
 import { WritingSystemData, WritingSystemScope } from '@entities/types/DataTypes';
-import ObjectSubtitle from '@entities/ui/ObjectSubtitle';
 import ObjectTitle from '@entities/ui/ObjectTitle';
 
+import CardField from '@shared/containers/CardField';
 import CommaSeparated from '@shared/ui/CommaSeparated';
 import CountOfPeople from '@shared/ui/CountOfPeople';
+import Deemphasized from '@shared/ui/Deemphasized';
 
 interface Props {
   writingSystem: WritingSystemData;
@@ -29,8 +36,9 @@ const WritingSystemCard: React.FC<Props> = ({ writingSystem }) => {
     scope,
     unicodeVersion,
   } = writingSystem;
-  const filterByScope = getScopeFilter();
   const { updatePageParams } = usePageParams();
+  const population =
+    populationUpperBound != null && populationUpperBound >= 100 ? populationUpperBound : null;
 
   return (
     <div>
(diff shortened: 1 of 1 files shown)


mchen11077:
Two quick questions before marking ready for review:

1. Should the Languages field be removed from WritingSystemCard since it’s not included in the new required field order?
2. Are the chosen icons (FileTextIcon, BlocksIcon, UsersIcon, ArrowLeftRightIcon, CodeIcon) appropriate?

src/entities/writingsystem/WritingSystemCard.tsx
```diff
@@ -42,13 +46,99 @@ const WritingSystemCard: React.FC<Props> = ({ writingSystem }) => {
...
-          <h4>Sample</h4>
-          {sample}
-        </div>
-      )}
-      {languages && Object.values(languages).length > 0 && (
+
+      <CardField
+        title="Sample"
+        icon={PencilLineIcon}
+        description="A Short example of how this writing system appears in text."
```
conradarcturus: "... appears in text" seems redundant since we are talking about writing.
```suggestion
        description="A single character from this writing system."
```

src/entities/writingsystem/WritingSystemCard.tsx
```diff
@@ -42,13 +46,99 @@ const WritingSystemCard: React.FC<Props> = ({ writingSystem }) => {
...
+                <Deemphasized>Variant of: </Deemphasized>
+                {parentWritingSystem ? (
+                  <HoverableObjectName object={parentWritingSystem} />
+                ) : (
+                  <Deemphasized>Unknown</Deemphasized>
+                )}
+              </div>
+            )}
+
+            {scope === WritingSystemScope.Group && (
```
conradarcturus: Only show this if it the group has items. It looks weird right now for Hani and Invented. If you want, you can add Hans & Hant to the columns "Group of other writing systems" for the row "Hani" in writingSystems.tsv. However, don't add anything to Invented since that will be too long.

```suggestion
            {scope === WritingSystemScope.Group && containsWritingSystems?.length && (
```

src/entities/writingsystem/WritingSystemCard.tsx
```diff
@@ -42,13 +46,99 @@ const WritingSystemCard: React.FC<Props> = ({ writingSystem }) => {
...
+          </div>
+        ) : (
+          <Deemphasized>Unknown</Deemphasized>
+        )}
+      </CardField>
+
+      <CardField
+        title="Population"
+        icon={UsersIcon}
+        description="This is a very rough estimate based on adding the populations that use the languages for this writing system, it's likely 50% to 400% off of the actual estimate."
```
conradarcturus: Thanks for using this from the `WritingSystemTable`. I realize now that I misused the word estimate.

I realize re-reading it now
```suggestion
        description="This is a very rough estimate based on adding the populations that use the languages for this writing system, it's likely 50% to 400% off of the actual population."
```

src/entities/writingsystem/WritingSystemCard.tsx
```diff
@@ -42,13 +46,99 @@ const WritingSystemCard: React.FC<Props> = ({ writingSystem }) => {
...
+        title="Text direction"
+        icon={ArrowLeftRightIcon}
+        description="The direction text is typically written and read for this writing system."
+      >
+        {rightToLeft == null ? (
+          <Deemphasized>Unknown</Deemphasized>
+        ) : (
+          <span>{rightToLeft ? 'Right to left' : 'Left to right'}</span>
+        )}
+      </CardField>
```
conradarcturus: This is what I asked for but now that I can see it I realize it doesn't work well. In a comparative view this looks identical -- someone reading this page won't quickly understand the significance of this.

Really, it matters when the language is written right to left (the usual case, one that UIs often fail). So let's get rid of this field and include it in the "Unicode support" section (since that's giving a technology overview).

So let's add it as an optional add-on for unicode support (I'll comment below how).

src/entities/writingsystem/WritingSystemCard.tsx
```diff
@@ -42,13 +46,99 @@ const WritingSystemCard: React.FC<Props> = ({ writingSystem }) => {
...
+        description="Whether this writing system is encoded in Unicode, which affects support across fonts, operating systems, and software."
+      >
+        {unicodeVersion === null ? (
+          <span>Not supported by Unicode</span>
+        ) : unicodeVersion != null ? (
+          <span>Unicode {String(unicodeVersion)}</span>
+        ) : (
+          <Deemphasized>Unknown</Deemphasized>
+        )}
+      </CardField>
```
conradarcturus: Add the Right-to-left warning here. Also note we're using a color defined in CSS (if we use "yellow" it would be too bright to see).

```suggestion
            {rightToLeft === true && (
              <>
                {' '}
                <Hoverable hoverContent="This writing system is written left to right, interfaces must handle this. Otherwise they risking disruptive display bugs for text in this writing system.">
                  <PilcrowLeftIcon color="var(--color-text-yellow)" size="1em" />
                </Hoverable>
              </>
            )}
      </CardField>
```

src/entities/writingsystem/WritingSystemCard.tsx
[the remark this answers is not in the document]
mchen11077: Removed.

conradarcturus (approved):
Thanks for making the changes :) There's 1 more update we can do but we can also clean that up in another round if you want to merge this change right now. Since I approved your change you can make the update and it will still be approved, so it is up to you.

PR #495, opened 2026-02-25
feat: add entity type tab navigation (#479)

### LINKED ISSUE #479: Add Horizontal Tab Navigation for Entity Types

Entity types (Languages, Locales, Census, Territory, Writing System, Variant Tag) are currently displayed as buttons in the left sidebar under the "Data" section, below  the search bar. This layout makes the sidebar feel cluttered and buries the primary navigation in a secondary panel.

Replace the current sidebar entity buttons with a horizontal tab bar positioned directly below the header, as the first element in DataPageBody. The tabs should include:

  - Language
  - Locale
  - Census
  - Territory
  - Writing System
  - Variant Tag

This gives entity type navigation a more prominent, discoverable position and follows a common UI pattern for switching between content categories.

### Required Changes
  - Create a new horizontal tab navigation component
  - Position it below the header, above the main content area (as the first element in DataPageBody)
  - Add an active state indicator (underline or the current theme's active state style)
  - Remove the entity type buttons from the left sidebar

### Visual Reference

|Current| Target|
|------|-------|
|Image | Image|

mchen11077:
Fixes #479.

截屏2026-02-24 下午5 45 16

Add a horizontal tab bar for entity type navigation below the search bar.

Changes:
- Add `EntityTypeTabs` component in `DataPageBody`
- Remove `ObjectTypeSelector` from `OptionsPanel`

### DIFF
--- a/src/widgets/pathnav/PathNav.tsx
+++ b/src/widgets/pathnav/PathNav.tsx
@@ -1,7 +1,7 @@
 import { SlashIcon } from 'lucide-react';
-import React, { useCallback } from 'react';
+import React from 'react';
 
-import { ObjectType, View } from '@features/params/PageParamTypes';
+import { View } from '@features/params/PageParamTypes';
 import Selector from '@features/params/ui/Selector';
 import {
   SelectorDisplay,
@@ -10,35 +10,13 @@ import {
 import usePageParams from '@features/params/usePageParams';
 import FilterPath from '@features/transforms/filtering/FilterPath';
 
-import ObjectTypeDescription from '@strings/ObjectTypeDescription';
-
 const PathNav: React.FC = () => {
   return (
     <PathContainer>
-      <ObjectTypeSelector />
       <SlashIcon size="1em" />
       <ViewSelector />
       <FilterPath />
     </PathContainer>
-  );
-};
-
-const ObjectTypeSelector: React.FC = () => {
-  const { objectType, updatePageParams } = usePageParams();
-  const goToObjectType = useCallback(
-    (objectType: ObjectType) => {
-      updatePageParams({ objectType });
-    },
-    [updatePageParams],
-  );
-
-  return (
-    <Selector
-      options={Object.values(ObjectType)}
-      onChange={goToObjectType}
-      selected={objectType}
-      getOptionDescription={(objectType) => <ObjectTypeDescription objectType={objectType} />}
-    />
   );
 };
 


src/pages/dataviews/EntityTypeTabs.tsx
```diff
@@ -0,0 +1,55 @@
...
+const EntityTypeTabs: React.FC = () => {
+  const { objectType, updatePageParams } = usePageParams();
+
+  return (
+    <div
+      style={{
+        display: 'flex',
+        marginBottom: '0.5em',
+        width: '100%',
+      }}
```
conradarcturus: too expanded, it can fit in 1 line `<div style={{ display: 'flex', marginBottom: '0.5em', width: '100%' }}>`

src/pages/dataviews/EntityTypeTabs.tsx
```diff
@@ -0,0 +1,55 @@
...
+              outline: 'none',
+              boxShadow: 'none',
+              borderBottom: isActive
+                ? '2px solid var(--color-button-primary)'
+                : '2px solid transparent',
+              cursor: 'pointer',
+              fontWeight: isActive ? 'bold' : 'normal',
+              color: isActive ? 'var(--color-button-primary)' : 'var(--color-text-secondary)',
+            }}
+          >
```
conradarcturus: We want to enable hover interactions as well -- but that's hard to do with react-alone. While we generally prefer not to use CSS, for hover interactions it is much better. We can add some class declarations to help the CSS rendering.

Also we can add hoverContent that shows a description

```suggestion
          <HoverableButton
            className={'tab' + (isActive ? ' active' : '')}
            hoverContent={/*... see the old ObjectTypeSelector code ...*/}
            key={type}
            onClick={() => updatePageParams({ objectType: type })}
            style={{ padding: '0.5em 1em' }}
          >
```

conradarcturus (commented):
While we prefer to avoid too much CSS (preferring the styling to be with the code using it), CSS handles hover & focus state coloring a little better. (Note I am not using color-button-hover since I think we may want to change that to the color mix as seen below)

You can add these lines to index.css, around where the other button styling is.

```
button.tab {
  background: none;
}

button.tab:hover,
button.tab:focus {
  background: color-mix(in srgb, var(--color-button-primary), var(--color-background) 90%);
}

button.tab:focus {
  border-color: var(--color-button-primary);
}
```

PR #500, opened 2026-02-28
refactor: move DataProvider to app level with deferred loading

### LINKED ISSUE #481: Make Data Context Available Earlier in Component Tree
### Problem

The data loading layer is currently scoped to the data page only. It is initialized when the user navigates to /data and loads the full dataset before rendering begins.
This tight coupling to a single page means that any component outside of the data page such as the header cannot access the loaded data.

This is a prerequisite for placing the search bar in the global header, but also a general improvement for any future feature that needs data access outside the data page.

### Proposed Solution

Move data initialization earlier in the component tree so it is available at the app level. To avoid increasing the app's initial load time, the data should be loaded lazily/deferred, meaning it starts loading early but does not block the initial render.

A future improvement could include a partial preloaded dataset for common anticipated queries (e.g. the most commonly searched languages), so results can appear before the full dataset finishes loading.

### Required Changes

  - Move the data provider higher in the component tree so it is accessible outside the data page
  - Ensure data loading remains deferred so it does not add blocking I/O costs before the app renders
  - Keep the bundle size impact minimal, the data loading code should not be included in the initial JS bundle

mchen11077:
Fixes #481 

Summary: Move DataProvider to App and keep it lazily loaded.

### Changes
- Mount PageParamsProvider and DataProvider in App
- Load providers via dynamic import in useEffect
- Remove providers from DataPage
 
### Test Plan
- `npm run dev`
- Check `/intro` renders immediately
- Check `/data` loads correctly
- `npm run build` and confirm DataProvider is a separate chunk

# Checklist
- [x] Data available globally
- [x] No blocking on initial render
- [x] DataProvider not in main bundle

### DIFF
--- a/src/entities/types/DataTypes.tsx
+++ b/src/entities/types/DataTypes.tsx
@@ -22,7 +22,7 @@ export interface ObjectBase {
   names: string[];
 }
 
-export type ObjectData =
+export type EntityData =
   | CensusData
   | LanguageData
   | LocaleData
@@ -30,6 +30,7 @@ export type ObjectData =
   | WritingSystemData
   | VariantTagData
   | KeyboardData;
+export type ObjectData = EntityData; // For now, all objects are entities
 export type ObjectDictionary = Record<string, ObjectData>;
 
 export enum WikipediaStatus {
--- a/src/features/data/context/useEntities.tsx
+++ b/src/features/data/context/useEntities.tsx
@@ -1,0 +1,51 @@
+import { useMemo } from 'react';
+
+import { useDataContext } from '@features/data/context/useDataContext';
+import { ObjectType } from '@features/params/PageParamTypes';
+import usePageParams from '@features/params/usePageParams';
+
+import { EntityData } from '@entities/types/DataTypes';
+
+const useEntities = (entityType?: ObjectType): EntityData[] => {
+  const { objectType: pageEntityType } = usePageParams();
+  const {
+    languagesInSelectedSource,
+    locales,
+    territories,
+    writingSystems,
+    variantTags,
+    censuses,
+    keyboards,
+  } = useDataContext();
+  const objects = useMemo(() => {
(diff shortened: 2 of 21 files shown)


src/app/App.tsx
```diff
@@ -9,12 +11,37 @@ import './index.css';
...
+    React.ComponentType<{ children: React.ReactNode }> | undefined
+  >(undefined);
+
+  React.useEffect(() => {
+    import('@features/params/PageParamsProvider').then((m) =>
+      setPageParamsProvider(() => m.default),
+    );
+    import('@features/data/context/DataProvider').then((m) => setDataProvider(() => m.default));
+  }, []);
+  if (!PageParamsProvider || !DataProvider) return null;
```
conradarcturus: We should return the children here. On a mid-tier mobile network the page takes 16s to load anything (way too long) then 4 more seconds to load PageParams & DataProvider classes (then a minute to load all of the data files). We should load the initial interface as fast as possible, we can load it 4 seconds early here if we return null.

I originally anticipated this task would be more difficult -- but it looks like the way that DataProvider runs it is totally fine to use this delayed import. Good job @mchen11077  finding a straightforward way to satisfy the needs.

Moving forward, we just need to keep an eye on page performance and broken experiences with partially loaded pages. Particular risks will be when we move data calls to use the external sql database and add new functionality (particularly visuals).

```suggestion
  if (!PageParamsProvider || !DataProvider) return children;
```

src/widgets/controls/selectors/ViewSelector.tsx
```diff
@@ -11,6 +11,7 @@ const ViewSelector: React.FC = () => {
   return (
     <Selector
       selectorLabel="Display"
+      selectorStyle={{ flexDirection: 'row', alignItems: 'center', gap: '0.25em' }}
```
conradarcturus: likewise, why the change?

PR #533, opened 2026-03-12
Population range filter(#446)

### LINKED ISSUE #446: Filters: Population Filter Min/Max Page options
For this change, can you add 2 page parameters that will allow people to filter the results shown by the population of the objects.

Steps
* Add to PageParamTypes PopulationLowerLimit and PopulationUpperLimit
* Add 2 Selectors (Can use LimitSelector as a template)
* Add to filtering logic (probably `filter.ts` but we're making some changes to this so it may look different when you get started)

You can follow this PR to see how a filter is usually added: https://github.com/Translation-Commons/lang-nav/pull/397

For now, just make it as 2 basic selectors, like the current LimitSelector. 
Image

Ultimately we want something fancier, like the screenshot below from the design. However, this will require more data & design work, so we'll reserve that for later.
Image

harsha19BCE0956:
Fixes #446

Summary:

This pull request implements the new Population Limits filter. It introduces the ability for users to filter results by establishing a minimum and/or maximum population bound, including new page parameters, a dedicated UI input selector.


conradarcturus (changes requested):
It is great to see so many things come together :) I left a note on the sections to update.

src/features/transforms/filtering/selectors/PopulationFilterSelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  // Local state to avoid input jumping or effect loops while typing
+  const [minInput, setMinInput] = useState<string | undefined>(undefined);
+  const [maxInput, setMaxInput] = useState<string | undefined>(undefined);
+
+  const displayMin = minInput !== undefined ? minInput : (populationLowerLimit ?? '');
+  const displayMax = maxInput !== undefined ? maxInput : (populationUpperLimit ?? '');
+
+  return (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '1em' }}>
```
conradarcturus: As I mentioned in the task, you should use the same code as `LimitSelector` to 1) use the same Selector element (similar styles) 2) show suggestions (eg. 0, 1, 10, 100, ...) 3) avoid custom implementations like you did here.

src/features/transforms/filtering/selectors/PopulationFilterSelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <span
+          style={{
+            fontSize: '0.9em',
+            color: 'var(--color-text-secondary)',
+            display: 'block',
+            marginBottom: '0.5em',
+          }}
+        >
+          Filter results to show only items with population above this threshold
+        </span>
```
conradarcturus: As you see, showing this text is unlike every single other selector on the website. Rather these are description fields that are seen via hovering over the info icon for regular selectors. It's `SelectorLabel`'s `description` field.

conradarcturus (changes requested):
Most things look good -- but the selector needs to be rewritten

conradarcturus (approved):
Some stylistic recommendations but they are low-priority. You can fix them now and still land it OR get back to it later.

conradarcturus:
Since Harsha disappeared, I resolved this with https://github.com/Translation-Commons/lang-nav/pull/599

PR #553, opened 2026-03-23
feat: move search bar into navbar (#482)

### LINKED ISSUE #482: Move Search Bar to Header

The search bar currently lives inside DataPageBody, rendered as the first element in the main content area, below the header/navbar. This makes it feel like a secondary element rather than the primary way to navigate the app.

Move the search bar into the header (navbar), centered between the site title on the left and the settings icon on the right, as shown in the Figma wireframe.

### Technical Blocker

The search bar currently sits inside the data page's component tree, which means it has access to the data context (loaded datasets, search index, etc.). The header/navbar is rendered above and outside of the data page entirely at the app root level alongside routing. Moving the search bar there would mean it no longer has access to the data context, so it would not function.

Before this change can be made, the data loading layer needs to be restructured so it is available at a higher level in the component tree (see #481 ).

### Required Changes
  - Move the searc#48h input into the header, centered between the title and settings icon
  - Decide whether to keep, remove, or integrate the "Search by" dropdown differently
  - Remove the search bar from its current position at the top of the main content area

Blocked by: #481 

### Visual Reference

|Current | Target|
|---------|---------|
|  Image | Image|

mchen11077:
Fixes #482 
Move the search bar from `DataPageBody` into the `NavBar`, centered between the nav links and Feedback button.

Changes:
- Add SearchBar to PageNavBar
- Remove SearchBar from DataPageBody  
- Integrate SearchBar and SearchBySelector into a single unified search input with shared white container
- Remove"Search by" text 
- Remove TextInput border and hide clear button when empty via CSS

截屏2026-03-23 下午1 52 46

### DIFF
--- a/src/features/transforms/search/SearchBar.tsx
+++ b/src/features/transforms/search/SearchBar.tsx
@@ -9,7 +9,6 @@ import {
 import TextInput from '@features/params/ui/TextInput';
 import usePageParams from '@features/params/usePageParams';
 
-import SearchBySelector from './SearchBySelector';
 import useSearchSuggestions from './useSearchSuggestions';
 
 const SearchBar: React.FC = () => {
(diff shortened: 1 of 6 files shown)


src/features/transforms/search/SearchBar.tsx
```diff
@@ -19,16 +19,26 @@ const SearchBar: React.FC = () => {
...
       >
         <TextInput
           label={<SearchIcon size="1em" display="block" style={{ padding: '0.5em' }} />}
-          inputStyle={{ minWidth: '20em', border: 'none' }}
+          inputStyle={{
+            minWidth: '12em',
+            flex: 1,
+            border: 'none',
+            backgroundColor: 'white',
+            color: '#04355c',
```
conradarcturus: I don't think you need flex here.

Also, again, use the CSS variables.

```suggestion
            border: 'none',
            backgroundColor: 'var(--color-background)',
            color: 'var(--color-text)',
```

conradarcturus (changes requested):
Thanks for working on this. I have a number of style recommendations though.

conradarcturus (approved):
Thanks for noticing the style problem -- we will need to move around some css components but I would not worry about that right now.

PR #563, opened 2026-03-24
Feature/code lookup

### LINKED ISSUE #442: Language Name <-> Language Code Conversion
The goal of this task is to make a tool in the LangNav website were people could paste line-separated language names and get line-separated language codes to speed up data intake.

For example, they'd enter

```
Urdu
Punjabi
Hindko
```

and it would produce

```
urd
lah/pan
lah/hind1271
```

As we define this task we can make more sub-issues. Some subtasks:
* Make basic matcher with 2 textareas (an input and output) that uses the same search functions as the SearchBar to convert input language names to the language code for the first result
* Show when there are multiple potential matches, maybe rank the matches more intelligently
* Let someone filter by country to only show matches to languages found in the country (or at least prioritize them)
* Dialect support since we usually don't match dialects -- maybe enable an option to see the branch of the family tree to confirm a result.
* Show composite languages names (like lah/pan, lah/hind1271) when macrolanguages are involved so both can be provided (if they apply).
* Fuzzy matching when words are spelled differently (eg. punjabi/panjabi)
* and whatever features you can think up.

uzairakbar:
Fixes #442

Summary: Adds a Code Lookup page to get language codes for a list of language names (one per line).

### Changes
- Added "Codes" page at `/code-lookup` in the nav bar
- Uses `findBestMatch` to score candidates by exact name > scope (macro vs. language) > code length
- And `buildCompositeCode` to produces `[macro]/[individual]` composite codes (e.g. `lah/pan`)

Future work: docs + cleanup +  improve matching logic (tbd)

### Test Plan and Screenshots
- `npx vitest run src/pages/code-lookup/__tests__/batchMatchCensus.test.ts` (census based accuracy eval for `findBestMatch`)
image

- Spin up `npm run dev`, navigate to `/code-lookup`, paste some names:
image

### DIFF
--- a/src/pages/code-lookup/__tests__/batchMatchCensus.test.ts
+++ b/src/pages/code-lookup/__tests__/batchMatchCensus.test.ts
@@ -1,0 +1,284 @@
+import { renderHook, waitFor } from '@testing-library/react';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+import * as DataProviderModule from '@features/data/context/DataProvider';
+const DataProvider = (DataProviderModule as any).DataProvider || (DataProviderModule as any).default;
+
+import * as UseDataContextModule from '@features/data/context/useDataContext';
+const useDataContext = (UseDataContextModule as any).useDataContext || (UseDataContextModule as any).default;
+
+import * as PageParamsModule from '@features/params/PageParamsProvider';
+const PageParamsProvider = (PageParamsModule as any).PageParamsProvider || (PageParamsModule as any).default;
+
+import { LanguageData } from '@entities/language/LanguageTypes';
+import { findBestMatch } from '../batchMatchLogic';
+
+
+const PUBLIC_DIR = path.resolve(__dirname, '../../../../public');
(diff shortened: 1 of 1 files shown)


src/pages/code-lookup/__tests__/batchMatchCensus.test.ts
```diff
@@ -0,0 +1,284 @@
+import { renderHook, waitFor } from '@testing-library/react';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+import * as DataProviderModule from '@features/data/context/DataProvider';
+const DataProvider = (DataProviderModule as any).DataProvider || (DataProviderModule as any).default;
```
conradarcturus: This is not how to mock things in a test environment. I think you want to do things like:

```
import usePageParams from '@features/params/usePageParams';

vi.mock('@features/params/usePageParams', () => ({
  default: vi.fn(),
}));

 (usePageParams as Mock).mockReturnValue(createMockUsePageParams());
```

conradarcturus (changes requested):
I left some comments but as said in the meeting -- let's develop this iteratively so my only ask for you is to fix the build/lint errors. Other features / styling can wait.

Run `npm run lint:fix` on your devserver to see what's wrong.

Issue #329, opened 2025-12-12
Language Proximity: Add languageMatch data


conradarcturus:
It would be fantastic to capture how close languages are.

One place to start is the CLDR information on closely related languages,  often used to inform consumers if they are missing a language and can fall back to another language's translations.
https://github.com/unicode-org/cldr/blob/main/common/supplemental/languageInfo.xml

PR #582, opened 2026-04-02
Language Proximity: add languageMatch data (#329)

preetsojitra2712:
## Summary
This PR implements issue #329 by importing CLDR `languageMatching.json` and adding `languageMatch` data to CLDR language objects.

It is intentionally scoped to the initial task:
- capture CLDR fallback language-match data
- avoid locale/script/region matches for now
- surface the results in Language Details

## What changed
- Import `languageMatching` from `cldr-core/supplemental/languageMatching.json` in `UnicodeData.tsx`
- Parse `supplemental.languageMatching[written-new].languageMatch`
- Attach per-language entries to `LanguageData.CLDR.languageMatch` with:
  - `desired`
  - `supported`
  - `distance`
  - `oneway`
- Restrict ingestion to **pure language-to-language** tags only (2-3 letter language codes), skipping locale/script/region/variable entries for this first step
- Skip self-matches (`en -> en`) to avoid noisy rows
- Add `languageMatch` to CLDR language type in `LanguageTypes.ts`
- Add display in `LanguageDetails` Connections section:
  - `Related Languages (CLDR)`
  - rendered as entries like `English (4 CLDR)`

## Files changed
- `src/features/data/load/supplemental/UnicodeData.tsx`
- `src/entities/language/LanguageTypes.ts`
- `src/entities/types/CLDRTypes.tsx`
- `src/widgets/details/LanguageDetails.tsx`
- `public/data/unicode/readme.md`

## Notes on scope
Follow-up annotation/report workflow code has been removed from this PR to keep it focused on the initial CLDR import + details display.

## Test plan
- [x] `npm run lint`
- [x] `npm run build`


src/widgets/reports/LanguageProximityAnnotationReport.tsx
```diff
@@ -0,0 +1,212 @@
...
+          </thead>
+          <tbody>
+            {visibleRows.map((row) => (
+              <tr key={row.id}>
+                <td>{row.desiredLanguage.nameDisplay + ' (' + row.desiredCode + ')'}</td>
+                <td>{row.supportedLanguage.nameDisplay + ' (' + row.supportedCode + ')'}</td>
+                <td>{row.match.distance}</td>
+                <td>{row.match.oneway ? 'one-way' : 'both'}</td>
+                <td>
+                  <button onClick={() => setMutualIntelligibility(row)}>
```
conradarcturus: use ToggleablePrediction

public/data/tc/language_match_annotations.tsv
```diff
@@ -0,0 +1 @@
+Desired	Supported	MutualIntelligibility	Bilingualism
```
conradarcturus: it would be good to have some examples

conradarcturus (changes requested):
Great! Thanks for making the change. In general you pivoted a bit too hard into my "ideas for future work" comment.

While you can work on the language_match_annotation tsv and report -- that really feels like a followup task, not 1 you should combine with this task. I don't want to ask you to delete the code (it's a good start) but it may prolong the time it takes to finish the initial task (adding the CLDR languageMatch data).

It would be better to have focused on the CLDR data import and then showing in LanguageDetails the results of the language match eg. in the Connections section showing "Related Languages: English (4<sup>CLDR</sup>), French (10<sup>CLDR</sup>)"

conradarcturus (approved):
Great! We will want to iterate on the design a bit but it looks like a solid start.

PR #628, opened 2026-05-06
view-settings implementation

### LINKED ISSUE #399: Settings: Move View Options to a settings modal

The sidebar (as of February 2026) is currently titled "Options" and contains 2 sections that serve very different purposes:
  - Common Actions: shortcuts and quick actions
  - Data: entity type selector buttons (Languages, Locales, Censuses, Territories, Writing Systems, Variant Tags) and source/profile options
  - Filter: the actual filter controls (Territory, Writing System, Language type, Vitality status, ISO status, etc.)
  - View: display controls (Cards/Table/Map/Hierarchy/Reports toggle, sort, color, scale options, brightness, etc.)

Mixing navigation, filtering, display preferences, and actions under one "Options" panel makes the sidebar cluttered and harder to understand at a glance.

This change is focused on moving "View" options outside of the options panel. Instead, use a `ViewModal` component like we do in `TableColumnSelector`.

The button to open this modal should be in the top-right of the screen next to the feedback button.

### Visual Reference

|Feb 2026 | Target|
|---------|---------|
|  Image | Image|

dhanush-kannan-dk:
Fixes #399

Summary: This PR implements view settings functionality, including a new modal interface and associated UI components. The changes enable users to configure view-related preferences through a dedicated settings panel.

### Changes

- User experience
  - Added a new navbar settings menu for quick access to view-related controls.
  - Introduced a compact top-right modal with open/close animations.
  - Moved view options out of the main options panel to keep filters focused and easier to navigate.
  - Added an animated light/dark theme toggle for page brightness customization.
  - Improved dropdown event handling to prevent race conditions during open/close interactions when using click-outside hook

- Data
  - No data changes.

- Refactors
  - Reorganized modal and settings-related UI logic for better maintainability.
  - Updated styling structure to support reusable compact overlay components.

### Screenshots

image

# Checklist

Feel free to check off or just delete items in this section as you have completed them.

## Summary

- [x] Mention the issue (#399)

## Testing

- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test`
  - [x] Include screenshots as noted below

### Visual changes

- [ ] Add screenshots to the table template at the top of this file. You can include images inside the table
  - [ ] Drag and drop images in the GitHub PR comment box to upload screenshots
- [x] Purely new views can just include the "after" screenshot.
- [ ] Since more views can be reproduced by just sharing the URL -- add links (eg. [link](https://translation-commons.github.io/lang-nav/data)) to the relevant page and/or conditions to reproduce the view.

### Data changes

- [x] No data changes

### Internal changes

- [x] Logical changes
- [x] Refactors, moving files around
- [x] If you notice any changes that require explanations, make sure to include the explanations in the code as well.

## Docs

- [x] Code is self-documenting, or if not, comments are added where needed.
- [ ] Updated markdown readme files documenting how the code behaves or how to develop in case there are any relevant changes to make.


src/widgets/controls/selectors/PageBrightnessSelector.tsx
```diff
@@ -1,23 +1,28 @@
...
 
-import Selector from '@features/params/ui/Selector';
-import usePageParams from '@features/params/usePageParams';
+import ThemeToggle from '@widgets/controls/selectors/ThemeToggle.tsx';
 
-import { PageBrightnessPreference } from '@shared/hooks/usePageBrightness';
-import { toSentenceCase } from '@shared/lib/stringUtils';
+import SelectorLabel from '@features/params/ui/SelectorLabel.tsx';
 
 const PageBrightnessSelector: React.FC = () => {
```
conradarcturus, line 7: How do you handle the "use device setting" option?

src/widgets/settings/Settings.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <>
+      <HoverableButton
+        className="primary"
+        onClick={() => setIsOpen(true)}
+        style={{ padding: '0.5em' }}
+      >
+        <SettingsIcon size="1.5em" display="block" />
+      </HoverableButton>
+      <div className={'view-settings'} style={{ fontSize: '0.9em' }}>
```
conradarcturus:
```suggestion
      <div className="view-settings" style={{ fontSize: '0.9em' }}>
```

src/widgets/controls/OptionsPanel.tsx
```diff
@@ -40,57 +27,47 @@ const OptionsPanel: React.FC = () => {
...
-          <SearchBySelector />
-          <ProfileSelector />
-        </OptionsPanelSection>
       </SelectorDisplayProvider>
     </ResizablePanel>
   );
 };
 
-const OptionsPanelSection: React.FC<
+export const OptionsPanelSection: React.FC<
```
conradarcturus: There is no need to use this component for toggling. In fact, your fullView or not status makes it better to just have 2 different containers, 1 for the filter panel and 1 for the settings modal.

src/features/params/ui/SelectorDropdown.tsx
```diff
@@ -39,10 +39,15 @@ export const SelectorDropdown: React.FC<React.PropsWithChildren<Props>> = ({
             role="listbox"
             style={{
               display: 'flex',
-              flexDirection: 'column',
-              width: 'fit-content',
+              flexDirection: 'row',
+              width: 'min-content',
               borderRadius: '1em',
-              backgroundColor: 'var(--color-background)',
+              backgroundColor: 'rgba(0,0,0,0)',
```
conradarcturus: if the background opacity is 0 just use transparent

src/features/params/ui/SelectorDropdown.tsx
```diff
@@ -39,10 +39,15 @@ export const SelectorDropdown: React.FC<React.PropsWithChildren<Props>> = ({
...
+              maxHeight: '50vh',
+              overflow: 'auto',
+              paddingRight: '1em',
+              direction: 'rtl',
```
conradarcturus: Why are you setting the direction to RTL? That's for languages like Arabic.

src/features/params/ui/SelectorDropdown.tsx
```diff
@@ -39,10 +39,15 @@ export const SelectorDropdown: React.FC<React.PropsWithChildren<Props>> = ({
             role="listbox"
             style={{
               display: 'flex',
-              flexDirection: 'column',
-              width: 'fit-content',
+              flexDirection: 'row',
```
conradarcturus: why is this row now?

src/features/layers/modal/modal.css
```diff
@@ -13,8 +13,31 @@
...
+}
+
+.overlay-top-right-open {
+  transform-origin: top right;
+  transform: scale(1);
+}
+
+.overlay-top-right-close {
+  transform-origin: top right;
+  transform: scale(0);
```
conradarcturus: oh cool! that's how you had it "pop up". Unfortunately I don't think the animation adds much value, and may even just be distracting (I'm happy for you to get a third opinion though and overrule me :) )

src/features/layers/modal/modal.css
```diff
@@ -13,8 +13,31 @@
...
 }
 
+.overlay-top-right {
+  width: auto;
+  height: auto;
+  border-radius: 10px;
+  align-items: flex-start;
+  justify-content: flex-end;
+  transition:
+    transform 0.8s linear(0, 0.45, 0.74, 0.88, 0.96, 1, 1.03, 1.05, 1.03, 1, 0.99, 1),
```
conradarcturus: The bouncing appears to be distracting. It's cute, but let's not rely too much on animations. I like the thought that went into it though.

src/shared/hooks/useClickOutside.tsx
```diff
@@ -17,3 +17,24 @@ export function useClickOutside(callback: () => void) {
   }, [callback]);
   return ref;
 }
+
+export function useClickOutsideTwo(callback: () => void) {
```
conradarcturus: Can you add a comment to the code explaining why we have a second useClickOutside? What is different about this one? We probably shouldn't call it "...Two" and instead give it a name that explains why it functions differently.

conradarcturus (changes requested):
Thanks for making this update -- sorry it took me awhile to get back to you, there was a lot going on and I didn't have time to test it out and understand what the code was doing.

I think it's generally pretty good but we should show it at the LangNav meeting on Tuesday and have other people weigh in on the animation & placement.

src/features/layers/modal/modal.css
dhanush-kannan-dk: Got it we shall revert 🤝

src/shared/hooks/useClickOutside.tsx
dhanush-kannan-dk: How about `useClickOutsideMultiRefs` or `useClickOutsideRefs` ?  

Basically I'm returning two-refs to handle dropdown event.
Earlier when we click the dropdown again, it was not closing. Usually, users would expect the dropdown to close when they click the dropdown-button again.

May be I can edit the Current `useClickOutside` function itself instead of adding a new one. What would you suggest?

src/features/params/ui/SelectorDropdown.tsx
dhanush-kannan-dk: It's more like a design suggestion. `flexDirection: 'row'` enables us to have a vertical scroll bar.
image

src/features/layers/modal/modal.css
dhanush-kannan-dk: Sure we can revert the animation after the review. Just submitted a suggestion : )

src/features/params/ui/SelectorDropdown.tsx
dhanush-kannan-dk: I noticed sometimes the scrollbar was getting cut off on the right, so I moved it to the left with `direction: rtl` to ensure the user has a clear visual cue that more options are available. We can add `direction: ltr` to the inner container to manage this. I'm happy to revert this if we find a better layout fix
image

src/widgets/controls/selectors/PageBrightnessSelector.tsx
dhanush-kannan-dk: As it is a classic toggle between Dark and Light mode, I didn't include an option for the follow-device. However, when the User open the website for the first time, the toggle will be set to either Dark or Light mode based on the device settings. Kindly check the below design for the third option (Auto)

Image

src/widgets/controls/OptionsPanel.tsx
dhanush-kannan-dk: Got it. Will create a separate container.

src/widgets/controls/selectors/PageBrightnessSelector.tsx
conradarcturus: As mentioned, use a computer icon for "follow device"

Thanks for iterating on this!

src/features/params/ui/SelectorDropdown.tsx
conradarcturus: Yea RTL means that the text is in RTL writing system like in Arabic or Hebrew. Don't use RTL for the wrong reason :p

Thanks for noticing it, there are few funny things with our dropdown formatting and (not for your change) we may want to redo dropdown styling.

src/features/params/ui/SelectorDropdown.tsx
conradarcturus: Gotcha gotcha -- so using flexDirection: row worked by only because it changed other things. You didn't realize what css properties you should have changed. Yes I agree enabling scrolling is a good idea -- but we can do it better.

Long term there are probably other updates we want but for now I recommend changing the styles passed into `Dropdown` to:

```
            style={{
              width: 'min-content',
              borderRadius: '1.2em',
              backgroundColor: 'var(--color-background)',
              maxHeight: '18em',
              overflow: 'auto',
              scrollbarWidth: 'thin',
            }}
```

conradarcturus:
I recommend using the same styles as the feedback modal.

In fact -- you may want to reuse the code for the feedback modal instead of the code for the center-page modal (that table columns settings is using right now).

Screenshot 2026-05-12 at 20 39 20

src/features/params/ui/Selector.tsx
conradarcturus, line 78: Ah, so you solution was side-stepping why the problem was happening. Thanks for working to fix it!

The prior code added a `useClickOutside` listener to `OptionsContainer` so when we click anywhere on the page it will close the `OptionsContainer`.

The problem is that the regular toggle in this `DropdownButton` is trying to toggle the state as well. So we have 2 `setExpanded` mutations happening at the same time. If 1 happens before the other they may interfere.

src/features/params/ui/Selector.tsx
```diff
@@ -40,17 +40,20 @@ function Selector<T extends React.Key>({
...
     <SelectorContainer manualStyle={selectorStyle} manualDisplay={display}>
       {selectorLabel && <SelectorLabel label={selectorLabel} description={selectorDescription} />}
 
-      <div style={{ position: 'relative', display: 'flex', alignItems: 'center' }}>
+      <div ref={triggerRef} style={{ position: 'relative', display: 'flex', alignItems: 'center' }}>
         {/* The dropdown menu or the button list */}
         <OptionsContainer
           isExpanded={expanded}
-          containerRef={optionsRef}
+          containerRef={containerRef}
```
conradarcturus: You don't need multiple refs. You can fix the bug with the existing `useClickOutside`

Just move the `containerRef` to the outer div (since that contains the `DropdownButton`) and remove the `OptionsContainer`'s attribute `containerRef`

dhanush-kannan-dk: Moving the ref to the outer wrapper won't work because the Dropdown uses createPortal. Since Portals render outside the parent DOM hierarchy, clicking an option is detected as an 'outside click.' This triggers the dropdown to close before the selection event can actually process.

Another workaround is adding a slight delay.
`const { optionsRef } = useClickOutside(() => {
    setTimeout(() => setExpanded(false), 100);
  });`
  It will fix the race condition.

src/features/params/ui/Selector.tsx
dhanush-kannan-dk: Yes yes. Exactly. That's why tried to return two refs from the function.

src/features/params/ui/SelectorDropdown.tsx
dhanush-kannan-dk: Got it. Since the flex property was already there, I tried tweaking it as a quick fix.
I ended up using a `borderRadius` of `1em` because `1.2em` was cutting off the inner container's border color

conradarcturus (approved):
Thanks for making the updates!

PR #641, opened 2026-05-19
Refactor map hover interactions

### LINKED ISSUE #448: Maps: Change hovercard interaction
## Problem

Right now there are some weaknesses with the hover interaction in the map component
* When you move your mouse over the screen the hovercards constantly disappear and reappear
* The hovercard takes up too much space on the map view
* It's hard to move your cursor on top of the hovercard.

Image

## Ideas

In the Figma, you can see some of the ideas we can do. One idea is to pin the infocards rather than have them disappear if the hover changes.

 https://www.figma.com/design/ZGJOgi5MnvtrKeaBJFrtZa/Language-Navigator?node-id=361-3&t=KCVkNbMb2Cpzaxi8-0
Image

To put this in practice I recommend a new scenario:
|Description|Screenshot|
|--|--|
|When hovering over a node on the map (a circle or a territory) it does not activate the hovercard component. Instead there's an independent div that shows the name of the language or country.|Image
|If you click on the circle or country, instead of opening the details panel, it will instead open the card. These cards should not use the same component as the current HoverCard -- rather it should just create an absolutely placed box centered below the click circle (in the screenshot its too the side since I did a quick mock, I think it would be better centered below). Note these card components should be independent so you could potentially have multiple cards shown on the screen at the same time.|Image
|You can still get to the details view by clicking on the expand button in the card.|Image

## How to do about this

In `MapHoverContent` remove the current text "click to...", limit the width, and add a floating button to open the details view. We should probably rename this to `MapCard` as well since it will no longer be triggered by a hover interaction.
```
<div style={{ position: 'relative', maxWidth: 200, fontSize: '0.75em' }}>
      {/* Click to see more information in the details panel. */}
      <div>
        <HoverableButton
          style={{
            position: 'absolute',
            top: 0,
            right: 0,
            padding: '0.25em',
          }}
        >
          <ExternalLinkIcon size="1.5em" display="block" onClick={() => updatePageParams({ objectID: object.ID })} />
        </HoverableButton>
      </div>
      <ObjectCard object={drawnObject} />
    </div>
```

In `MapCircles` and `MapTerritories` change how `useHoverCard` is used -- instead of rendering `getHoverContent`, render a smaller component with just the `object.nameDisplay` and "Click for more" (or whatever text you like).

Then the `<circle>`s onClick event should go from updating the page parameter to instead create a div on the page (ideally centered below the circle) with the results from `getHoverContent` (which we probably should rename `getHoverContent` to `getCard` while we are at it). Now you can click on multiple circles and view multiple things. I leave the process of creating these temporarily divs up to you (maybe create a FloatingCard component?). Do some testing to see if there are any regressions in the UX but don't think you need to necessarily solve everything, we can iterate.

Thanks for working on this task!

mchen11077:
Fixes #448  

截屏2026-05-18 下午10 43 13

#### Changes
- Replaced `MapHoverContent` with `MapCard`
- Added click-to-open floating cards
- Moved navigation into expand button
- Simplified hover preview content
- Clicking outside the floating card now closes it


src/features/map/ObjectMap.tsx
```diff
@@ -53,15 +61,35 @@ const ObjectMap: React.FC<Props> = ({ objects, maxWidth = 2000 }) => {
...
+      const rect = containerRef.current?.getBoundingClientRect();
+      if (!rect) return;
+
+      setFloatingCards((prev) => [
+        ...prev,
+        { id: `${object.ID}-${Date.now()}`, object, x: clientX - rect.left, y: clientY - rect.top },
+      ]);
+    },
+    [containerRef],
+  );
```
hdprajwal, line 94: `openCard` always appends and uses `Date.now()` as part of the key, so clicking the same circle twice pins two identical cards. Dedupe by `object.ID` and use it as the key.

hdprajwal, line 94: The cards don't follow zoom/pan. After any zoom/pan the underlying circle moves and the card where they were opened. 
Possible fix - Render the floating cards inside contentRef and position them with map coordinates so they move with the rest of the map.

src/features/map/ObjectMap.tsx
```diff
@@ -53,15 +61,35 @@ const ObjectMap: React.FC<Props> = ({ objects, maxWidth = 2000 }) => {
...
+
+  useEffect(() => {
+    setFloatingCards([]);
+  }, [objectType, objects]);
+
```
hdprajwal, line 118: This `useEffect` resets pinned cards whenever the parent re-creates the objects array, even though the user's pinned cards may still be valid. 
No need to fix this now, You can create another issue for this to be fixed later.

src/features/map/ObjectMap.tsx
```diff
@@ -92,21 +121,43 @@ const ObjectMap: React.FC<Props> = ({ objects, maxWidth = 2000 }) => {
...
             drawableObjects={drawableObjects}
-            getHoverContent={getHoverContent}
+            getCard={getCard}
+            openCard={openCard}
             scalar={1200 / maxWidth}
             coloringFunctions={coloringFunctions}
           />
         </div>
+
+        {floatingCards.map((card) => (
```
hdprajwal: Currently only "click outside" closes pinned cards, and clicking another circle adds a new card. We can add a small `×` icon in the card corner to let users close individual cards rather than all at once. 
This is a suggestion, can be another issue along with the useEffect fix.

hdprajwal (changes requested):
Nice work, this is a solid improvement over the old behavior and the structure is clean.
Have left inline comments. Happy to clarify any of them, just reply on the thread.

src/features/map/ObjectMap.tsx
mchen11077: Thanks! I've narrowed the dependency to `objectType` in this PR so recreating the `objects` array no longer clears still-valid pinned cards. The broader ID-based pruning cleanup is tracked in #642 as a follow-up.

hdprajwal (approved):
Nice work, this looks good to me! and can be merged

PR #653, opened 2026-06-09
Map: fix node size during zoom

### LINKED ISSUE #636: Maps: Scale circles & text better
Map circles and text scale to the width of the map but they are not necessarily the right size. When you zoom in they also don't change. It would be good to:

* Experiment with the best base size for circles and text
* Re-adjust node size when zooming in (maybe linearly or maybe at square root rate)

See the inconsistencies easily when looking at the macrolanguage map and looking at the maps shown by Chinese
https://translation-commons.github.io/lang-nav/data?view=Map&fieldFocus=Name&languageScopes=4&objectID=zho

Image

mchen11077:
Fixes #636 
## Summary
- Added an onZoom callback to useMapZoom that exposes the zoom factor relative to the initial map fit.
- Applied scale(1 / zoomFactor) to node groups 
- Tuned circle radius and stroke width, and hid text labels below 1.5× zoom to reduce clutter in the full-world view.

<p align="center">
  
  
</p>

### DIFF
--- a/src/features/map/MapCentroids.tsx
+++ b/src/features/map/MapCentroids.tsx
@@ -19,6 +19,7 @@ type Props = {
   drawableObjects: DrawableData[];
   openCard: (obj: DrawableData, x: number, y: number) => void;
   scalar: number;
+  zoomFactor: number;
   coloringFunctions: ColoringFunctions;
 };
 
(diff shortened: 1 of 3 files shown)


src/features/map/MapCentroids.tsx
```diff
@@ -140,7 +146,8 @@ const Circle: React.FC<NodeProps> = ({
   onMouseLeave,
 }) => (
   <circle
-    r={scale + 1}
+    r={scale + 1.5}
+    strokeWidth={0.4}
```
conradarcturus, line 150: not to be solved today -- but we should figure out a good strokeWidth

conradarcturus (approved):
Fantastic. It's fun to play around with too.

Image

PR #667, opened 2026-06-15
Sort: Replace sort selector with popup card containing sort options

### LINKED ISSUE #579: Change sort to a card
The current button to change the sorting field works well but there is a bit more sorting functionality hidden away in the view options that we can bring up into that interactive component. Namely, the `SortDirection` and the `SecondarySortBy` options.

Instead of adding more visual noise, let's organize it differently and replace the current `SortBySelector` shown as a piece of text (with similar styling as now) that is a `HoverableButton`. The `hoverContent` is a card with the 3 selectors for the 3 settings, and maybe some explanatory text.

Replace the dropdown interaction here:
Image

with a menu that shows the 3 sorting options currently showing in the View Options at the top right corner of the page.

Image

After you do this, you can remove the sorting options from the view options.

mchen11077:
Fixes #579

截屏2026-06-15 上午11 44 55

## Summary
- Replace the toolbar sort dropdown with a `SortPopupCard` that contains Sort By, Secondary Sort By, and Sort Direction.
- Remove the corresponding sort controls from View Settings.
- Remove the unused `showLabel` prop from `SortBySelector`.

### DIFF
--- a/src/features/transforms/sorting/SortPopupCard.tsx
+++ b/src/features/transforms/sorting/SortPopupCard.tsx
@@ -1,0 +1,39 @@
+import { ArrowDownUpIcon } from 'lucide-react';
+import React from 'react';
+
+import PopupCard from '@features/layers/popupcard/PopupCard';
+import { SelectorDisplay } from '@features/params/ui/SelectorDisplayContext';
+import { getOptionStyle } from '@features/params/ui/SelectorOption';
+import usePageParams from '@features/params/usePageParams';
+
+import { PositionInGroup } from '@shared/lib/PositionInGroup';
+
+import SecondarySortBySelector from './SecondarySortBySelector';
+import SortBySelector from './SortBySelector';
+import SortDirectionSelector from './SortDirectionSelector';
+
+const SortPopupCard: React.FC = () => {
+  const { sortBy } = usePageParams();
+
+  return (
+    <div className="selector" style={{ gap: '0.25em' }}>
+      <ArrowDownUpIcon size="1.2em" />
+      <PopupCard
+        buttonLabel={<>{sortBy} ▶</>}
+        buttonClassName="selected"
+        buttonStyle={getOptionStyle(SelectorDisplay.Dropdown, true, PositionInGroup.Standalone)}
+        description="Change how items are sorted."
+        title="Sort"
+        body={() => (
+          <div style={{ display: 'flex', flexDirection: 'column', gap: '0.5em' }}>
+            <SortBySelector />
+            <SecondarySortBySelector />
+            <SortDirectionSelector />
+          </div>
+        )}
+      />
+    </div>
+  );
+};
+
+export default SortPopupCard;
--- a/src/features/transforms/sorting/SortBySelector.tsx
+++ b/src/features/transforms/sorting/SortBySelector.tsx
@@ -1,4 +1,3 @@
-import { ArrowDownUpIcon } from 'lucide-react';
 import React from 'react';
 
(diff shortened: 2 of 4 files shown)


src/features/transforms/sorting/SortPopupCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <PopupCard
+        buttonLabel={<>{sortBy} ▶</>}
+        buttonClassName="selected"
+        buttonStyle={getOptionStyle(SelectorDisplay.Dropdown, true, PositionInGroup.Standalone)}
+        description="Change how items are sorted."
+        title="Sort"
+        body={() => (
+          <div style={{ display: 'flex', flexDirection: 'column', gap: '0.5em' }}>
+            <SortBySelector />
+            <SecondarySortBySelector />
```
conradarcturus, line 30: The label right now is "Secondary Sort By" do you think the label "Tie breaker" would be easier to understand? Otherwise we can keep it as-is.

src/features/transforms/sorting/SortPopupCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <div className="selector" style={{ gap: '0.25em' }}>
+      <ArrowDownUpIcon size="1.2em" />
+      <PopupCard
+        buttonLabel={<>{sortBy} ▶</>}
+        buttonClassName="selected"
+        buttonStyle={getOptionStyle(SelectorDisplay.Dropdown, true, PositionInGroup.Standalone)}
+        description="Change how items are sorted."
+        title="Sort"
+        body={() => (
+          <div style={{ display: 'flex', flexDirection: 'column', gap: '0.5em' }}>
```
conradarcturus, line 28: Let's keep the selectors in line, add

```
              width: 'max-content',
              alignItems: 'flex-end',
```

src/features/transforms/sorting/SortPopupCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import SecondarySortBySelector from './SecondarySortBySelector';
+import SortBySelector from './SortBySelector';
+import SortDirectionSelector from './SortDirectionSelector';
+
+const SortPopupCard: React.FC = () => {
+  const { sortBy } = usePageParams();
+
+  return (
+    <div className="selector" style={{ gap: '0.25em' }}>
```
conradarcturus, line 19: Add `flexWrap: 'nowrap'`

src/features/transforms/sorting/SortPopupCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <ArrowDownUpIcon size="1.2em" />
+      <PopupCard
+        buttonLabel={<>{sortBy} ▶</>}
+        buttonClassName="selected"
```
conradarcturus, line 23: (I may be wrong) The semantic category you want here maybe is "primary". Selected should be only used when there are multiple options.
```suggestion
        buttonClassName="primary"
```

src/features/transforms/sorting/SortPopupCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+        buttonStyle={getOptionStyle(SelectorDisplay.Dropdown, true, PositionInGroup.Standalone)}
```
conradarcturus, line 24: Interesting use! makes sense. Thanks for re-using styles.

conradarcturus (approved):
Some minor edits but otherwise lgtm.

Do you think we should add any explanations to the component or is it okay just as selectors?

PR #673, opened 2026-06-16
feat(map): complete pinned floating card restore from URL

### LINKED ISSUE #658: Pins: Add URL parameter
Add a new URL parameter `pinned` that we will use to store the currently selected/pinned entities.

You'll want to update a number of entries in [PageParamTypes.tsx](https://github.com/Translation-Commons/lang-nav/blob/409304470a2735ef3b99cacceee93f300ec603f4/src/features/params/PageParamTypes.tsx#L57) -- PageParamKey, PageParams, PageParamsOptional

You'll notice many other files that will show errors that you should update, check how other ones are handled. This new parameter will be an array of strings (see how other arrays are handled).

Now of course we cannot test the new parameter without adding it somewhere -- let's update the Map pins to use this new parameter! See the new floating cards that @mchen11077  added in https://github.com/Translation-Commons/lang-nav/pull/641 -- use this new parameter to control which floating cards should appear (so update add/remove `pinned` IDs using usePageParams and have `floatingCards` be built using `useMemo` depending on the pinned IDs.

akash-belide:
Fixes #658 

Summary: This finishes work started in the prior commit (feat(params): add pinned URL param for map floating cards). That earlier commit added a pinned URL parameter and synced it when opening/closing cards, but floating cards still lived in local React state. After a refresh, the URL could contain pinned IDs, but the cards would not reappear.

This PR makes pinned the single source of truth for which floating cards are open. Floating cards are now derived from the pinned page parameter, so pinned cards can be restored from the URL after refresh.

### Changes
- User experience
  - Restores pinned floating cards after a page refresh when the URL contains pinned entity IDs.
  - Allows shared URLs with pinned values to reopen the same floating cards.
  - Clears pinned cards from the URL when clicking the map background.
- Logical changes
  - Removed local React state as the source for floating cards.
  - Derived floatingCards from the pinned page parameter using useMemo.
  - Updated openCard/closeCard to only add/remove entity IDs to pinned.
  - Added object type change handling that clears pinned only when the object type actually changes.
- Refactors
  - Moved Robinson map scale values into shared constants.

### Screenshots
image
image

### DIFF
--- a/src/features/map/EntityMap.tsx
+++ b/src/features/map/EntityMap.tsx
@@ -44,7 +44,7 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
     onZoom: setZoomFactor,
   });
 
-  const { colorBy, objectType } = usePageParams();
+  const { colorBy, objectType, pinned, updatePageParams } = usePageParams();
   const [floatingCards, setFloatingCards] = useState<FloatingCard[]>([]);
   const [mapScale, setMapScale] = useState(1);
 
@@ -79,6 +79,7 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
 
   const openCard = useCallback(
     (entity: DrawableData, clientX: number, clientY: number) => {
+      updatePageParams({ pinned: [...pinned, entity.ID] });
       const rect = contentRef.current?.getBoundingClientRect();
       if (!rect) return;
 
@@ -93,12 +94,16 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
         { id: entity.ID, entity, x, y },
       ]);
     },
-    [contentRef, mapHeight],
+    [contentRef, mapHeight, pinned],
   );
 
-  const closeCard = useCallback((id: string) => {
-    setFloatingCards((prev) => prev.filter((card) => card.id !== id));
-  }, []);
+  const closeCard = useCallback(
+    (id: string) => {
+      updatePageParams({ pinned: pinned.filter((pin) => pin != id) });
+      setFloatingCards((prev) => prev.filter((card) => card.id !== id));
+    },
+    [pinned],
+  );
 
   const handleZoomIn = useCallback(() => {
     zoomIn();
--- a/src/features/map/EntityMap.tsx
+++ b/src/features/map/EntityMap.tsx
@@ -108,7 +108,7 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
       if (pinned.includes(entity.ID)) return;
       updatePageParams({ pinned: [...pinned, entity.ID] });
     },
-    [pinned, updateMapScale,updatePageParams],
+    [pinned, updateMapScale, updatePageParams],
   );
 
   const closeCard = useCallback(
--- a/src/features/map/EntityMap.tsx
+++ b/src/features/map/EntityMap.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import { ObjectType } from '@features/params/PageParamTypes';
 import usePageParams from '@features/params/usePageParams';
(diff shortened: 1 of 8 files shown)


src/features/map/EntityMap.tsx
```diff
@@ -77,28 +83,40 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
-
-      const scale = rect.width / MAP_INTERNAL_WIDTH;
-      setMapScale(scale);
-
-      const x = ((clientX - rect.left) / rect.width) * MAP_INTERNAL_WIDTH;
-      const y = ((clientY - rect.top) / rect.height) * mapHeight;
+  // Floating cards are derived from the pinned page param so they can be fully restored from the
+  // URL after a refresh. Each card is positioned at its entity's Robinson centroid.
+  const floatingCards = useMemo<FloatingCard[]>(() => {
+    const drawableById = new Map(drawableEntities.map((entity) => [entity.ID, entity]));
```
conradarcturus, line 88: We have a function for this but your code is also effective so its up to you

```suggestion
    const drawableById = groupBy(drawableEntities, ent => ent.ID);
```

src/features/map/EntityMap.tsx
```diff
@@ -77,28 +83,40 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
+      .map((id) => {
+        const entity = drawableById.get(id);
+        if (entity == null || entity.latitude == null || entity.longitude == null) return undefined;
+
+        const { x: robinsonX, y: robinsonY } = getRobinsonCoordinatesShifted(entity);
+        return {
+          id,
+          entity,
+          x: MAP_INTERNAL_WIDTH / 2 + robinsonX * MAP_ROBINSON_X_SCALE,
+          y: mapHeight / 2 - robinsonY * MAP_ROBINSON_Y_SCALE,
```
conradarcturus, line 99: Ah, I wasn't sure if we needed to do this math. Thanks for figuring it out.

src/features/map/EntityMap.tsx
```diff
@@ -77,28 +83,40 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
 
   const coloringFunctions = useColors({ objects: drawableEntities });
 
-  const openCard = useCallback(
-    (entity: DrawableData, clientX: number, clientY: number) => {
-      const rect = contentRef.current?.getBoundingClientRect();
-      if (!rect) return;
-
-      const scale = rect.width / MAP_INTERNAL_WIDTH;
-      setMapScale(scale);
```
conradarcturus, line 86: Ah, you didn't keep this in the new function, we'll need it to verify the scale when we add new pins.

src/features/map/EntityMap.tsx
```diff
@@ -77,28 +83,40 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
+      .filter((card): card is FloatingCard => card != null);
+  }, [pinned, drawableEntities, mapHeight]);
 
-      setFloatingCards((prev) => [
-        ...prev.filter((card) => card.id !== entity.ID),
-        { id: entity.ID, entity, x, y },
-      ]);
+  const openCard = useCallback(
+    (entity: DrawableData) => {
+      if (pinned.includes(entity.ID)) return;
```
conradarcturus, line 108: The scale may not be correct, add a scale check here.
```suggestion
      updateMapScale();
      if (pinned.includes(entity.ID)) return;
```

src/features/map/EntityMap.tsx
```diff
@@ -115,9 +133,19 @@ const ObjectMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
 
+  // Restore the map scale on mount so URL-restored cards are sized correctly before any zoom.
   useEffect(() => {
-    setFloatingCards([]);
-  }, [objectType]);
+    requestAnimationFrame(updateMapScale);
+  }, [updateMapScale]);
+
+  // Clear pinned cards when the object type changes, but not on the initial mount so that pinned
+  // cards can still be restored from the URL after a refresh.
```
conradarcturus: Hehe, thanks for adding this, it is a good idea.

Instead of writing the code here, write it in `clearContextDependentParams` inside the block when the entity type changes `if (newParams.objectType !== undefined && newParams.objectType !== prevOrDefault.objectType) {` -- you can just add the line `next.delete('pinned');`

conradarcturus (approved):
Instead of opening a new PR/changing branches you can amend your previous PR/branch

Nonetheless it works. Thanks for making it fully interactive. I'm accepting the PR so you can update it and then merge it when you are ready (rather than waiting for another review).

PR #682, opened 2026-06-19
feat: add card pinning and keep pinned results at top

### LINKED ISSUE #659: Pins: Add to CardList UI
When someone hovers over a card there should be a floating icon button (use the Lucide icon `PinIcon`) that when clicked on will add the pin to the page param "pinned" (see usePageParams) -- the icon should not be visible when the hover is inactive && the pin is off. When hovering over the pin if it is already clicked then it should should the PinOffIcon. You probably want to use `HoverableButton` to add hovercontext text "Pin <item name> to the page".

Update `useFilteredObjects` to not filter things that are pinned. So say, for instance, you search for Nahuatl and pin one of the results then change the search to Mapuche then Mapuche stays there.

akash-belide:
Fixes #659 

Summary: Adds card pinning from the card list and ensures pinned entities stay visible at the top of filtered results.

### Changes

- User experience
  - Added a floating pin button to cards
  - The pin button appears when the card is hovered or when the card is already pinned.
  - Added hover text for **Pin to the page** and **Unpin from the page**.
  - Pinned cards keep the pin button visible.
  - Hovering over the pin button on a pinned card switches the icon from PinIcon to PinOffIcon.
  
- Logical changes
  - Pinned entities bypass active filters.
  - Pinned entities remain visible even when the current search or filters would normally hide them.
  - Pinned entities are sorted to the front
  - Pinned entities preserve the order stored in the pinned page param.
  - Non-pinned entities continue to follow the normal sort order.
  

### Screenshots
- Pinned
image

- Pinned entities bypass active filters
image

- Hover text for **Pin to the page** and **Unpin from the page**
image
image

### DIFF
--- a/src/features/transforms/filtering/useFilteredEntities.tsx
+++ b/src/features/transforms/filtering/useFilteredEntities.tsx
@@ -1,6 +1,7 @@
 import { useMemo } from 'react';
 
 import useEntities from '@features/data/context/useEntities';
+import usePageParams from '@features/params/usePageParams';
 import { getSortFunction } from '@features/transforms/sorting/sort';
 
 import { ObjectData } from '@entities/types/DataTypes';
@@ -31,6 +32,7 @@ const useFilteredEntities = <T extends ObjectData>({
 }: UseFilteredEntitiesParams<T>): { filteredEntities: T[]; allEntities: T[] } => {
   // Implementation of filtering logic goes here
   const pageEntities = useEntities() as T[]; // Get all objects of the relevant type from context
+  const { pinned } = usePageParams();
   // TODO use useFilters for all of these
   const filters = useFilters();
   const filterByScope = useScope ? getScopeFilter() : () => true;
@@ -45,15 +47,28 @@ const useFilteredEntities = <T extends ObjectData>({
     return allEntities
       .filter(
         (obj) =>
-          filterByScope(obj) &&
-          filterBySubstring(obj) &&
-          filterByConnections(obj) &&
-          filterByVitality(obj) &&
-          filterByPopulation(obj),
+          // Pinned entities always remain visible regardless of the active filters.
+          pinned.includes(obj.ID) ||
+          (filterByScope(obj) &&
+            filterBySubstring(obj) &&
+            filterByConnections(obj) &&
+            filterByVitality(obj) &&
+            filterByPopulation(obj)),
       )
-      .sort(sortFunction);
+      .sort((a, b) => {
+        const aIndex = pinned.indexOf(a.ID);
+        const bIndex = pinned.indexOf(b.ID);
+        const aPinned = aIndex !== -1;
+        const bPinned = bIndex !== -1;
+        // Pinned entities always come first, in the order they were pinned, so they
+        // reliably stay at the top (and on the first page) across refreshes.
+        if (aPinned && bPinned) return aIndex - bIndex;
+        if (aPinned !== bPinned) return aPinned ? -1 : 1;
+        return sortFunction(a, b);
+      });
   }, [
     allEntities,
+    pinned,
     filterByScope,
     filterBySubstring,
     filterByConnections,
--- a/src/widgets/cardlists/CardInCardList.tsx
+++ b/src/widgets/cardlists/CardInCardList.tsx
@@ -1,10 +1,12 @@
-import { PinIcon, PinOffIcon } from 'lucide-react';
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback } from 'react';
 
-import HoverableButton from '@features/layers/hovercard/HoverableButton';
 import usePageParams from '@features/params/usePageParams';
 
 import { ObjectData } from '@entities/types/DataTypes';
+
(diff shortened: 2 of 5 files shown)


src/features/transforms/filtering/useFilteredEntities.tsx
```diff
@@ -45,15 +47,28 @@ const useFilteredEntities = <T extends ObjectData>({
...
       )
-      .sort(sortFunction);
+      .sort((a, b) => {
+        const aIndex = pinned.indexOf(a.ID);
+        const bIndex = pinned.indexOf(b.ID);
+        const aPinned = aIndex !== -1;
+        const bPinned = bIndex !== -1;
+        // Pinned entities always come first, in the order they were pinned, so they
+        // reliably stay at the top (and on the first page) across refreshes.
+        if (aPinned && bPinned) return aIndex - bIndex;
```
conradarcturus, line 65: oh nice, I didn't expect this but it makes sense to use the pin array to decide the order of the pins. Originally I had thought it would stay in sorted order amongst the pins, but this makes sense too.

src/widgets/cardlists/CardInCardList.tsx
```diff
@@ -67,6 +83,32 @@ const CardInCardList: React.FC<Props> = ({ children, getBackgroundColor, object
       }}
       tabIndex={0}
     >
+      {showPinButton && (
+        <HoverableButton
```
conradarcturus: This is a lot of markup, it would probably be best to move this to its own class `CardPinButton`

src/widgets/cardlists/CardInCardList.tsx
```diff
@@ -67,6 +83,32 @@ const CardInCardList: React.FC<Props> = ({ children, getBackgroundColor, object
...
+            border: 'none',
+            padding: '0.25rem',
+            lineHeight: 0,
+            color: isPinned ? 'var(--color-button-primary)' : 'var(--color-text)',
+          }}
+        >
+          <span
+            aria-label={isPinned ? 'Unpin from the page' : 'Pin to the page'}
+            onMouseEnter={() => setIsHoveringPin(true)}
+            onMouseLeave={() => setIsHoveringPin(false)}
```
conradarcturus: We should probably all these as parameters to HoverableButton, but..

For this button it may be better to be completely different and actually not use `HoverableButton`. One advantage of the other hoverable buttons is that uses CSS hover to color the button. but since you hardcoded the background we cannot switch the background color to the usual hover state (blue bg white foreground). The background is nice because it gives a wider click area. I do agree that given its role when its not in a hover state it is better transparent.

So 1 options 1) change the background color & foreground color when `isHovering` or 2) move most of the styling to CSS. You already gave it a class name, you can also append a class when its pinned (eg. `className={'CardPinButton' + (isPinned ? ' pinned' : '')}`.

src/widgets/cardlists/CardInCardList.tsx
```diff
@@ -11,8 +13,21 @@ interface Props {
...
 
 const CardInCardList: React.FC<Props> = ({ children, getBackgroundColor, object }) => {
-  const { updatePageParams, objectID } = usePageParams();
+  const { updatePageParams, objectID, pinned } = usePageParams();
   const [isHovering, setIsHovering] = React.useState(false);
+  const [isHoveringPin, setIsHoveringPin] = React.useState(false);
+
+  const isPinned = pinned.includes(object.ID);
+  const togglePin = useCallback(() => {
+    updatePageParams({
```
conradarcturus, line 22: If the card moves it stays in the hover state which is odd (however if the card did not move then this is wrong).

Hmmm, I think we may want to control colors with CSS -- I mentioned later that you can either use isHoveringPin to make more fine-grained controls but I notice that this doesn't fix all the rendering bugs (cards stay hovered even when it moves away from the cursor because the mouseLeave event did not trigger).

I think the css `:hover` attribute is more accurate than what we are doing now.

```suggestion
    setIsHoveringPin(false);
    updatePageParams({
```

conradarcturus (approved):
It functions well!

I left some comments but they aren't blocking -- I'll leave it to your judgment.

akash-belide:
Extracted the pin button from CardInCardList into CardPinButton and move styling to cardListStyles.css. Use CSS :hover for visibility and icon swapping so hover state stays correct when cards reorder, and override global button:focus style.

src/widgets/cardlists/CardInCardList.tsx
```diff
@@ -1,23 +1,29 @@
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback } from 'react';
 
 import usePageParams from '@features/params/usePageParams';
 
 import { ObjectData } from '@entities/types/DataTypes';
 
+import CardPinButton from './CardPinButton';
+
+import './cardListStyles.css';
```
conradarcturus, line 9: great job!

conradarcturus (approved):
Looks great! I made some updates to how we handle hover icons but we'll want to do some followup later -- let's add this feature first.

PR #683, opened 2026-06-23
Pins: Add sidebar for pinned map cards

### LINKED ISSUE #669: Pins: Reconsider Map Pins
Now that we're moving the Map Cards to use pins we need to ask what the best design for them is.

On one hand we can leave them static on the map, but on the other they take up too much real estate. Maybe we could make a card list on the Maps page of the pinned entities. For this task can you design the next version of this visualization, present your ideas, then we implement your ideas.

RutanshS:
Fixes #669 

Summary: Replaces the floating map cards with a sidebar panel. Pinned entities appear in a scrollable sidebar on the left side of the map. Hovering on the said cards highlights the corresponding map centroids or territories.

### Changes

- User experience
  - Clicking a map centroid or territory pins/unpins it. (Toggle)
  - Pinned entities appear in a sidebar panel instead of floating over the map.
  - Hovering a sidebar card highlights the corresponding centroid or territory on the map.
  - Clicking anywhere else on the map no longer closes all cards.
- Logical changes
  - EntityMap: Replaced FloatingCard state and map positioning logic with the MapSidebar Component and uses the pinnedEntities derived from the URL.
  - MapCentroids / MapTerritories: Added hoveredId and pinnedIds for visual feedback on pinned entities. Fixed pinCard typing.
- Refactors
  - Renamed openCard to pinCard and closeCard to unpinCard to match the actual behavior.
  - Removed the mapScale, updateMapScale and handleZoomIn function wrappers that existed only for positioning floating cards.

Out of scope/Future work: Unit Tests, Clear All Button in sidebar, Maximum Pin Limit (Maybe). 

#### Pinned Single Card
Pinned Single Card

#### Pinned Multiple Cards
Pinned Multiple Cards

#### Pinned Card Hover
Pinned Card Hover


RutanshS:
### Follow Up: Theme Relevant Border Colors
Latest commit addresses the feedback that the pinned entity border colors do not appear to be contrasting based on the theme. Fixed it to have different contrasting colors for the pinned centroid or territory borders based on the theme. The chosen colors are from the existing design guideline document. Also restored the map border.

src/features/map/EntityMap.tsx
```diff
@@ -32,17 +26,12 @@ type Props = {
   maxWidth?: number;
 };
```
conradarcturus, line 28: Maps can be in the main data view panel but also in the Details panel for Languages and Territories (click on a language and scroll to the bottom of the details and you'll see maps). It's weird to have the siderbar in those places because the page width is too small. Can you add a parameter `allowSidebar` default set to false and in `ViewMap.tsx` only set it to true.

https://translation-commons.github.io/lang-nav/data?view=Map&pinned=bel&objectID=bel

Image

src/features/map/EntityMap.tsx
```diff
@@ -83,135 +64,99 @@ const EntityMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
           aspectRatio: MAP_ASPECT_RATIO,
+          display: 'flex',
           overflow: 'hidden',
-          cursor: 'grab',
-          position: 'relative',
+          background: 'var(--color-background)',
         }}
-        onClick={() => updatePageParams({ pinned: [] })}
       >
+        <MapSidebar
```
conradarcturus: only show this if `allowSidebar` is true

src/features/map/EntityMap.tsx
```diff
@@ -83,135 +64,99 @@ const EntityMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
 
   const coloringFunctions = useColors({ objects: drawableEntities });
 
-  // Floating cards are derived from the pinned page param so they can be fully restored from the
-  // URL after a refresh. Each card is positioned at its entity's Robinson centroid.
-  const floatingCards = useMemo<FloatingCard[]>(() => {
+  const pinnedEntities = useMemo(() => {
```
conradarcturus: Since this is only used in `MapSidebar` I would remove this code from `EntityMap` and pass `drawableEntities` to `MapSidebar` and do this memo there.

src/features/map/EntityMap.tsx
```diff
@@ -83,135 +64,99 @@ const EntityMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
+  const pinCard = useCallback(
     (entity: DrawableData) => {
-      updateMapScale();
-      if (pinned.includes(entity.ID)) return;
-      updatePageParams({ pinned: [...pinned, entity.ID] });
+      if (pinned.includes(entity.ID)) {
+        updatePageParams({ pinned: pinned.filter((id) => id !== entity.ID) });
+      } else {
+        updatePageParams({ pinned: [...pinned, entity.ID] });
+      }
```
conradarcturus: Given the above changes, let's change this method from always being `pinCard` to instead be called `onClick`. If the sidebar is allowed, then update pins, if the sidebar is not allowed then let's just open/or update the details page: `updatePageParams({objectID: entity.ID})`

src/features/map/EntityMap.tsx
```diff
@@ -83,135 +64,99 @@ const EntityMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
+          border: '1px solid var(--color-text-secondary)',
           width: '100%',
           aspectRatio: MAP_ASPECT_RATIO,
+          display: 'flex',
           overflow: 'hidden',
-          cursor: 'grab',
-          position: 'relative',
+          background: 'var(--color-background)',
         }}
-        onClick={() => updatePageParams({ pinned: [] })}
```
conradarcturus, line 174: yeap, good idea. We no longer need to dismiss all pins with an off-card click.

src/features/map/EntityMap.tsx
```diff
@@ -83,135 +64,99 @@ const EntityMap: React.FC<Props> = ({ entities, maxWidth = 2000 }) => {
...
-        zoomOut={handleZoomOut}
-        resetTransform={handleResetTransform}
-      />
+      <ZoomControls zoomIn={zoomIn} zoomOut={zoomOut} resetTransform={resetTransform} />
 
       <div
-        ref={containerRef}
         style={{
-          border: '1px solid var(--border-color)',
+          border: '1px solid var(--color-text-secondary)',
```
conradarcturus: Ah, this was a mistake I made when I used a color that does not exist.

However, tbh, I think we are better with no border at all.

src/features/map/MapCentroids.tsx
```diff
@@ -17,18 +17,22 @@ import './map.css';
 
 type Props = {
   drawableEntities: DrawableData[];
-  openCard: (obj: DrawableData, x: number, y: number) => void;
+  pinCard: (obj: DrawableData) => void;
```
conradarcturus: Do `onClick` instead

src/features/map/MapCentroids.tsx
conradarcturus: if the sidebar is available label this "Pin to sidebar" if it is not "Open in details panel"

src/features/map/MapCentroids.tsx
```diff
@@ -140,18 +152,33 @@ const Circle: React.FC<NodeProps> = ({
...
+          ? 'var(--color-button-primary)'
+          : color == null
+            ? 'var(--color-button-primary)'
+            : 'transparent'
+    }
+    style={
+      isHovered
+        ? { filter: 'brightness(1.2)', transition: 'all 0.15s ease-in-out' }
+        : { transition: 'all 0.15s ease-in-out' }
+    }
```
conradarcturus: There is a CSS file with some styles `map.css` -- if we need to do math or add overrides let's keep it here (eg. `r`, `fill`) but for others like `scale`, `strokeWidth`, `filter`, `transition` let's update that CSS. This seems like good values for `map.css`. This also helps keeps the mouse hover and sidebar hover consistent.

```
circle {
  transition:
    fill 0.25s,
    stroke 0.25s,
    scale 0.25s;
  pointer-events: fill;
  cursor: pointer;
  stroke-width: 1px;
}

circle:hover,
circle.hovered {
  filter: brightness(1.2);
  scale: 1.5;
  stroke: var(--color-button-primary);
}

circle.pinned {
  stroke: var(--color-button-primary);
}
```

```suggestion
  <circle
    className={(isHovered ? 'hovered' : '') + (isPinned ? ' pinned' : '')}
    r={scale + 1.5}
    fill={color ?? 'transparent'}
    stroke={color == null ? 'var(--color-button-primary)' : 'transparent'}
```

src/features/map/MapTerritories.tsx
```diff
@@ -56,13 +60,27 @@ const MapTerritories: React.FC<Props> = ({
...
         element.style.fill = '#bcbcbc';
       }
 
       element.style.cursor = 'pointer';
+
+      if (hoveredId === territory.ID) {
+        element.style.opacity = '0.7';
+      } else {
+        element.style.opacity = '1';
+      }
```
conradarcturus: Way too much manual styling, let's use classes instead (that's a way to sync the hover states as well)
```suggestion
} else {
element.style.fill = '#bcbcbcbc';
}
```

src/features/map/MapTerritories.tsx
```diff
@@ -56,13 +60,27 @@ const MapTerritories: React.FC<Props> = ({
...
     });
-  }, [territories, getColor, isTerritoryInList, colorBy, svgLoaded]);
+  }, [territories, getColor, isTerritoryInList, colorBy, svgLoaded, hoveredId, pinnedIds]);
```
conradarcturus, line 83: Add a separate `useEffect` here to update class names.

and to `map.css` add definitions:
```
.mapTerritory {
  cursor: pointer;
}

.mapTerritory:hover,
.mapTerritory.hovered {
  filter: brightness(1.2);
  stroke: var(--color-button-primary);
  stroke-width: 2px;
}

.mapTerritory.pinned {
  stroke: var(--color-button-primary);
  filter: brightness(1.1);
  stroke-width: 2px;
}
```


```suggestion
  // Manage hovered and pinned states
  useEffect(() => {
    if (!svgLoaded) return;

    forEachTerritory((territory, element) => {
      element.classList.add('mapTerritory');
      element.classList.remove('hovered');
      element.classList.remove('pinned');
      element.classList.remove('inList');
      if (pinnedIds.includes(territory.ID)) element.classList.add('pinned');
      if (hoveredId === territory.ID) element.classList.add('hovered');
    });
  }, [svgLoaded, hoveredId, pinnedIds]);
```

src/features/map/MapTerritories.tsx
conradarcturus: Let's remove the manual styling here and use CSS instead as mentioned before. Also changing from opacity to brightness.
```suggestion
(territory: TerritoryData) => (ev: MouseEvent) => {
showHoverCard(
<div>
<strong>{territory.nameDisplay}</strong>
<div style={{ color: 'var(--color-text-secondary)' }}>Click for more</div>
</div>,
ev.clientX,
ev.clientY,
);
```

conradarcturus, line 103: We can simplify this now that we have css

src/features/map/MapSidebar.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <div
+      style={{
+        width: '300px',
+        height: '100%',
+        overflowY: 'auto',
+        borderRight: '1px solid var(--color-text-secondary)',
+        display: 'flex',
+        flexDirection: 'column',
+        gap: '1em',
+        padding: '1em',
```
conradarcturus: I don't think the border is necessary

```suggestion
        display: 'flex',
        flexDirection: 'column',
        gap: '1em',
        padding: '1em',
```

src/features/map/MapSidebar.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type Props = {
+  pinnedEntities: DrawableData[];
+  objectType: ObjectType;
+  onClose: (id: string) => void;
+  hoveredId: string | null;
+  setHoveredId: (id: string | null) => void;
+};
+
+const MapSidebar: React.FC<Props> = ({
```
conradarcturus, line 17: I have other ideas on the appearance of this but I don't want to tackle too many changes at a time ;) So for now I'll call this good and we'll iterate on it after the existing feedback.

conradarcturus (changes requested):
This is looking very promising! I have a number of recommendations on styling and triggering events but it's a good start.

conradarcturus (approved):
Thanks for making all of the changes! There are probably some updates we want to do to the placement of the sidebar but I think that's a good followup for later. Let's land this code now :)

PR #651, opened 2026-06-05
Fix Combined language hierarchy for ISO-only groupings

whwan4570:
## Summary
- Add `languageFamilyCombinedOverrides.tsv` and loader to adjust **Combined** parent links when ISO has family groupings Glottolog lacks
- Fixes Albanian (`sqj` → `sqi`), Quechuan (`qwe` → `que`), Basque (`euq` → `eus`), and Formosan/Austronesian (`map` → `fox`, with `poz` remaining under `map`)

Fixes #650

## Changes
- `public/data/tc/languageFamilyCombinedOverrides.tsv`
- `src/features/data/load/extra_entities/CombinedFamilyOverrides.tsx`
- `src/features/data/load/extra_entities/__tests__/CombinedFamilyOverrides.test.ts`
- `src/features/data/load/CoreData.tsx` — apply overrides after Glottolog load

## Test plan
- [x] Verified Combined + Hierarchy UI locally for Albanian, Quechuan, Basque, Formosan
- [x] `npm run test:run` for `CombinedFamilyOverrides.test.ts`

PR #684, opened 2026-06-23
Fix Combined language hierarchy for ISO-only groupings

### LINKED ISSUE #650: Improve language hierarchy when ISO has a record that Glottolog does not
Regularly Glottolog is a superset of ISO so we don't have this issue. But there are a few places where ISO has a grouping that is not in Glottolog. When people pick a data authority of ISO or Glottolog things are okay -- but when people select a data authority of Combined we have to handle the parent/child relationship.

## Examples

For example, see Formosan [fox] -- in ISO it's a grouping under Austronesian [map] but Glottolog does not use it (in part because its a regional grouping not a genetic grouping).

https://translation-commons.github.io/lang-nav/data?languageScopes=4%2C3%2C2%2C5&searchString=formosan&objectID=map

Image

All of the children of Austronesian [map] should be in the Formosan [fox] group (except Malayo-Polynesian [poz]). However Glottolog does not have a Formosan grouping.

All known groupings:
* Austronesian [map] -> Formosan [fox] -> most current descendents of Austronesian except Malayo-Polynesian [poz]
* Quechuan [qwe] should contain Quechuan Macrolanguage [que]
* Albanian [sqj] should contain Albanian Macrolanguage [sqi]
* Basque [euq] should contain Basque Language [eus] <-- note this isn't currently broken but it will be better defined in this override

There may be others! You can try finding them

## Tips for Solution

`public/data/tc/glottocodeToISO.tsv` has our list of manual glottocodes and most parent/child relations come from the glottolog import `public/data/glottolog/glottolog.tsv`, ISO family data `public/data/tc/familiesToLanguages.tsv` (should be prefixed with iso since its an TC file but its specific to ISO only) or our main language list `public/data/tc/languages.tsv`.

However none of those data files accomplish what we are trying to achieve -- combining these hierarchies. So we should make a new manual data file in the TC folder, let's say `public/data/tc/languageFamilyCombinedOverrides.tsv`

Load this data into LangNav during the CoreData step and add overrides (after the ISO and Glottolog loads) then test out the hierarchy view with the known groupings to see if it it is fixed.

whwan4570:
## Summary

Rebased on latest `master` and addresses @conradarcturus review on #651.

- Replace `reparent-children-except` with explicit row-per-child entries in `languageFamilyCombinedOverrides.tsv`
- Add Formosan children under `fox`, plus Conrad's additional ISO-only groupings (`sdv`, `omv`, `sai`, `ssa`, `nic`)
- Apply overrides after Glottolog load; only `Combined.parentLanguageCode` is modified

Fixes #650

Supersedes #651 (same change set, rebased on current master).

## Test plan

- [x] `npm run test:run` — `CombinedFamilyOverrides.test.ts`
- [x] `npm run lint` && `npm run build`
- [x] Hierarchy UI verified locally (Combined source) — see PR comment with screenshots

Made with [Cursor](https://cursor.com)

### DIFF
--- a/public/data/tc/languageFamilyCombinedOverrides.tsv
+++ b/public/data/tc/languageFamilyCombinedOverrides.tsv
@@ -1,0 +1,25 @@
+# Manual overrides for Combined language family hierarchy.
+# ISO-only groupings that Glottolog does not use need explicit Combined parent/child fixes.
+# Only Combined.parentLanguageCode is modified; ISO and Glottolog sources are unchanged.
+parentLanguageCode	childLanguageCode	notes
+sqj	sqi	Albanian macrolanguage in Albanian family
+qwe	que	Quechuan macrolanguage in Quechuan family
+euq	eus	Basque language in Basque family
+map	fox	Formosan in Austronesian
+fox	east2493	East Formosan in Formosan
+fox	atay1246	Atayalic in Formosan
+fox	pwn	Paiwan in Formosan
+fox	bnn	Bunun in Formosan
+fox	dru	Rukai in Formosan
+fox	nort2899	Northwest Formosan in Formosan
+fox	tsou1250	Tsouic in Formosan
+fox	pyu	Puyuma in Formosan
+fox	west2572	Western Plains Austronesian in Formosan
+sdv	nilo1247	Nilotic in Eastern Sudanic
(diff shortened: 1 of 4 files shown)


whwan4570:
Superseded by #685 — branch moved to upstream `Translation-Commons/lang-nav` so deploy-preview can run with repo secrets.

PR #688, opened 2026-06-25
fix: scale colorbar tick font size

### LINKED ISSUE #679: Map Colorbar: Calibrate the # of ticks to be shown based on the pixel width
When we color nodes on a map we show a colorbar. However that text on the colorbar gets too dense when the map is minimized.

2 Possible approaches:
1. Change the font size based on the colorbar width
2. Change the number of tickets shown. `Colorbar.tsx` has the function `getTicks` that determines. The function `getIdealNumberOfTicks` is hard-coded for a few types but it works best for view-page maps, we should decrease that number when the colorbar width is smaller.

|State|Screenshot|
|--|--|
|Zoomed Out|Image
|Zoomed In|Image

dhanush-kannan-dk:
Fixes #679

Summary: The colorbar below the map had a fixed font size and number format
regardless of container width, causing tick labels to overlap on narrow
viewports.

### Changes

- User experience
  - Colorbar tick labels now scale font size continuously based on container
    width (0.5em at ≤450px → 1em at ≥900px), preventing overlap when the
    map is in a narrow layout
  - Colorbar number labels switch to short notation (K, M, B) below 700px
    wide, replacing long labels like "190 thousand" with "190K"
- Logical changes
  - `ResizeObserver` added to `ColorBar` to track its own pixel width
- Refactors
  - Fixed a pre-existing React hooks violation in `ColorBar` where `useMemo`
    was called after an early `return null`
  - Fixed a locale-sensitive test in `InteractiveEntityTable.test.tsx` that
    hardcoded `'1,234,567'` — replaced with `toLocaleString()` so it passes
    on machines using non-Western number formatting (e.g. Indian lakh system)

Out of scope/Future work: Could further reduce the number of ticks shown at
very narrow widths if font scaling alone is not sufficient.

### Test Plan and Screenshots

How to test the changes in this PR: Run `npm run dev`, navigate to a map
view, enable Color By (e.g. Population), then resize the browser window
narrower to observe font size and number format adapting.

## Testing

- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test`
  - [x] Test updated for one failed test case
  - [x] Include screenshots as noted below

image

### DIFF
--- a/src/features/map/ZoomControls.tsx
+++ b/src/features/map/ZoomControls.tsx
@@ -17,29 +17,31 @@ const ZoomControls: React.FC<Props> = ({
   resetTransform,
   containerWidth = 800,
 }) => {
-  const iconSize = containerWidth < 500 ? 16 : containerWidth < 700 ? 20 : 24;
-  const padding = containerWidth < 500 ? '0.25em' : containerWidth < 700 ? '0.35em' : '0.5em';
-  const buttonStyle: React.CSSProperties = {
-    padding,
-    display: 'flex',
-    alignItems: 'center',
-    justifyContent: 'center',
-  };
+  const fontSize = containerWidth < 500 ? '0.7em' : containerWidth < 700 ? '0.85em' : '1em';
   return (
-    <div style={containerStyle}>
+    <div style={{ ...containerStyle, fontSize }}>
       <HoverableButton onClick={zoomIn} style={buttonStyle} aria-label="Zoom in">
-        <ZoomInIcon size={iconSize} />
+        <ZoomInIcon style={iconStyle} />
       </HoverableButton>
 
       <HoverableButton onClick={zoomOut} style={buttonStyle} aria-label="Zoom out">
-        <ZoomOutIcon size={iconSize} />
+        <ZoomOutIcon style={iconStyle} />
       </HoverableButton>
 
       <HoverableButton onClick={resetTransform} style={buttonStyle} aria-label="Reset">
-        <Maximize2Icon size={iconSize} />
+        <Maximize2Icon style={iconStyle} />
       </HoverableButton>
     </div>
   );
+};
+
+const iconStyle: React.CSSProperties = { width: '1em', height: '1em' };
+
+const buttonStyle: React.CSSProperties = {
+  padding: '0.5em',
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
 };
 
 const containerStyle: React.CSSProperties = {
--- a/src/features/map/ZoomControls.tsx
+++ b/src/features/map/ZoomControls.tsx
@@ -8,21 +8,30 @@ type Props = {
   zoomIn: () => void;
   zoomOut: () => void;
   resetTransform: () => void;
+  containerWidth?: number;
 };
 
-const ZoomControls: React.FC<Props> = ({ zoomIn, zoomOut, resetTransform }) => {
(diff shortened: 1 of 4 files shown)


src/features/transforms/coloring/ColorBar.tsx
```diff
@@ -37,25 +37,48 @@ type Props = {
...
+    [coloringFunctions, containerWidth],
+  );
   const renormalize = isFieldWholeNumbersOnly(coloringFunctions.colorBy)
     ? (value: number) =>
         coloringFunctions.getNormalizedValue(
           Math.round(coloringFunctions.getDenormalizedValue(value)),
         )
     : undefined;
 
+  if (minValue === undefined || maxValue === undefined) {
```
conradarcturus, line 65: Thanks for fixing the react problem by moving this after the hooks.

src/features/transforms/coloring/ColorBar.tsx
```diff
@@ -37,25 +37,48 @@ type Props = {
 const ColorBar: React.FC<Props> = ({ coloringFunctions }) => {
   const { minValue, maxValue } = coloringFunctions;
   const { colorGradient } = usePageParams();
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [containerWidth, setContainerWidth] = useState(800);
```
conradarcturus: nit: Call this `colorBarRef` and `colorBarWidth` -- container is too generic.

src/features/transforms/coloring/ColorBar.tsx
```diff
@@ -173,7 +199,7 @@ function getTicks(coloringFunctions: ColoringFunctions): { position: number; lab
 
   const formatter = new Intl.NumberFormat(undefined, {
     notation: 'compact',
-    compactDisplay: 'long', // or 'long' for “thousand”, “million”
+    compactDisplay: widthPx < 700 ? 'short' : 'long',
```
conradarcturus, line 202: nice call

src/features/map/ZoomControls.tsx
```diff
@@ -8,21 +8,35 @@ type Props = {
...
+  const iconSize = containerWidth < 500 ? 16 : containerWidth < 700 ? 20 : 24;
+  const padding = containerWidth < 500 ? '0.25em' : containerWidth < 700 ? '0.35em' : '0.5em';
+  const buttonStyle: React.CSSProperties = {
+    padding,
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+  };
   return (
     <div style={containerStyle}>
```
conradarcturus, line 15: Instead of these padding & icon breaks, since padding and icon uses relative units (`em`) -- why not just change the container's font size?

Your approach works but it requires passing in a lot of styles when CSS already is configured to use relative sizing.

conradarcturus (approved):
Great :) Changes look good and it shows that you have an eye for detail. I have some stylistic notes but you'll welcome to ignore them if you prefer your choices.

src/features/transforms/coloring/ColorBar.tsx
dhanush-kannan-dk: Yes yes... I agree

src/features/map/ZoomControls.tsx
dhanush-kannan-dk: Sure... Makes sense 🤝. Seems Master branch have some changes related to this. So undid the changes I did for the zoom icons.

PR #694, opened 2026-06-27
Add pin column to all tables via shared PinButton

### LINKED ISSUE #661: Pins: Add to Table UI
Let's add a new column on the far-left that will allow people to pin rows to the table.

Add a new column that will appear in all tables -- `CommonColumns.tsx` is a good place for this. This new column should show the `PinIcon` with similar interactions as mentioned in the other task #659 

In order to insert this column to all table views, look at `useColumnVisibility.tsx` -- in the inner function of `const visibleColumns = useMemo(`...`() => columns.filter((column) => columnVisibility[column.key]),` -- you can append the column using the syntax `() => [PinColumn, ...columns.filter((column) => columnVisibility[column.key])]`

akash-belide:
Fixes #661 

Summary: Add a far-left pin column to all tables so users can pin rows via the existing pinned page param, using the same CSS-driven PinButton interactions as cards.

Changes
User experience
- Every table gets a far-left pin column.
- Pin button appears on row hover; stays visible when the row is already pinned (blue icon).
- Hovering a pinned button swaps PinIcon → PinOffIcon to signal unpin.
- Tooltips: "Pin to the page" / "Unpin from the page".
- Pin column is the sticky left column on horizontal scroll.

Logical changes

- TablePinCell toggles entity IDs in the pinned page param on click.
- useColumnVisibility prepends PinColumn to visibleColumns for every table, independent of column visibility settings.

Refactors

- Extracted shared PinButton (src/shared/ui/PinButton.tsx + pinButton.css) for reuse by cards and tables.
- Added row-hover visibility rules in tableStyles.css 
- Updated InteractiveEntityTable tests for +1 always-present column header.

### Screenshots
image

image


src/shared/ui/PinButton.tsx
```diff
@@ -3,26 +3,29 @@ import React from 'react';
...
+
 interface Props {
+  className?: string;
   isPinned: boolean;
   onTogglePin: () => void;
 }
 
-const CardPinButton: React.FC<Props> = ({ isPinned, onTogglePin }) => {
-  // The action on a pinned card is always to unpin it, so the label reflects that.
+const PinButton: React.FC<Props> = ({ className, isPinned, onTogglePin }) => {
```
conradarcturus, line 14: great idea to generalize the pin button :)

src/features/table/tableStyles.css
```diff
@@ -29,3 +29,14 @@ summary:hover {
...
+/* The pin column button is only revealed when the row is hovered or the row is
+   pinned. The button's appearance lives in pinButton.css. */
+tbody tr .PinButton {
+  visibility: hidden;
+}
+
+tbody tr:hover .PinButton,
+tbody tr .PinButton.pinned {
+  visibility: visible;
+}
```
conradarcturus, line 42: btw `BaseEntityTable` has styling for the first column that makes the ID column sticky so if you see lots of columns and scroll to the right the ID column stays. However that rule is for `idx === 0` but now that is the Pin column. Can you update it so its `idx <= 1`?

While you are there, you can also fix a fun. When I added it, I did not think about conflicting styles so the background row highlight doesn't work for this column. So I would do something similar to before when we add a new value (eg. `alwaysVisible`) to `className` if `idx <= 1` and do the `backgroundColor` and etc in this css file.

src/features/table/useColumnVisibility.tsx
```diff
@@ -35,7 +36,10 @@ function useColumnVisibility<T extends ObjectData>(
   );
 
   const visibleColumns = useMemo(
```
conradarcturus, line 38: perfect

src/features/table/CommonColumns.tsx
```diff
@@ -1,14 +1,38 @@
...
+  }, [isPinned, pinned, object.ID, updatePageParams]);
+
+  return <PinButton isPinned={isPinned} onTogglePin={togglePin} />;
+};
+
+export const PinColumn: TableColumn<ObjectData> = {
+  key: 'Pin',
+  label: '',
+  render: (object) => <TablePinCell object={object} />,
+  exportValue: () => '',
```
conradarcturus, line 33: Ah! I realize since it is always visible, it will be part of the table export. You should update `TableExport` to omit the "Pin" column when `pinned` is empty.

src/widgets/cardlists/cardListStyles.css
```diff
@@ -17,49 +17,17 @@
   border-color: var(--color-button-primary);
 }
 
-.CardPinButton {
+/* Position the shared pin button in the card's top-right corner. The button's
+   own appearance (colors, icon swap) lives in pinButton.css. */
+.CardInCardList .PinButton {
```
conradarcturus, line 22: great!

conradarcturus (approved):
This looks fantastic :D

I realized there are a few interaction problems but I'll mark this approved now because you have a good idea about what you are doing so you can make the update and land the new code without waiting for another approval.

akash-belide:
Pin column polish: sticky columns, hover fix, and smarter export.

- The pin column and first data column now stay frozen on the left while scrolling horizontally (sticky styling moved to `.alwaysVisible` in tableStyles.css).
- Fixed the row-hover highlight that was being hidden behind the sticky columns' opaque background.
- Table export now omits the pin column when nothing is pinned, and marks pinned rows as "Pinned" when pins exist.

src/features/table/BaseEntityTable.tsx
```diff
@@ -42,25 +42,25 @@ function BaseEntityTable<T extends ObjectData>({ visibleColumns, objects }: Prop
...
+                // row-hover highlight live in tableStyles.css under `.alwaysVisible`.
+                const isSticky = idx <= 1;
+                return (
+                  <td
+                    key={column.key}
+                    className={isSticky ? `${valueType} alwaysVisible` : valueType}
+                    style={{
+                      maxWidth: MAX_COLUMN_WIDTH,
+                      padding: '0.25em 0.5em',
+                    }}
```
conradarcturus: minor: can 1-line this
```suggestion
                    style={{ maxWidth: MAX_COLUMN_WIDTH, padding: '0.25em 0.5em' }}
```

src/features/table/tableStyles.css
```diff
@@ -29,3 +29,41 @@ summary:hover {
...
+
+/* The first two body columns (the pin column and the first data column) stay
+   pinned to the left edge while the table is scrolled horizontally. */
+td.alwaysVisible {
+  /* Pin column width, reused as the left offset of the second sticky column. */
+  --pin-column-width: 2.75em;
+  position: sticky;
+  background-color: var(--color-background);
+  z-index: 5; /* ZIndex.TableStickyColumn */
```
conradarcturus, line 51: I love the commenting

src/features/table/tableStyles.css
```diff
@@ -29,3 +29,41 @@ summary:hover {
...
+}
+
+td.alwaysVisible:nth-child(2) {
+  left: var(--pin-column-width);
+}
+
+/* Sticky cells paint an opaque background that would otherwise hide the row
+   hover highlight, so re-apply the highlight to them while the row is hovered. */
+tbody tr:hover td.alwaysVisible {
+  background-color: var(--color-background-primary);
```
conradarcturus: `color-background-primary` does not exist. -- but I realize we don't have a color for what you need here. I would have recommended `color-background-hover` but that  `#007bff1f` -- which uses transparency (allows us to use nested highlighting). Can you add `color-background-hover-opaque` to colors.css? The hex value you want is `#e6f4fe`. Add a version for dark mode too (On my Mac I use Digital Color Meter to measure the hex value).

conradarcturus (approved):
Thanks for making the update for the export :) 1 comment about the coloring but otherwise gtg

PR #720, opened 2026-07-17
Auto zoom

### LINKED ISSUE #678: Maps: Automatically zoom into coordinates of visible objects
EntityMap starts zoomed out -- but often this is too far out, especially if we're just seeing a few entities like the Sino-Tibetan languages

https://translation-commons.github.io/lang-nav/data?view=Map&languageFamilyFilter=Sino-Tibetan+%5Bsit%5D

|State|Screenshot|
|--|--|
|Zoomed Out|Image
|Zoomed In|Image

Can you update EntityMap so that when the map loads (and/or maybe when entities change) it changes the zoom to appropriately be sized for the objects inside?

akash-belide:
Fixes #678

Summary: The map (`EntityMap`) always loaded fully zoomed out to show the entire world, which was often too far out when only a few entities were displayed, for example when filtering to a single language family like Sino-Tibetan. This PR makes the map automatically zoom and center to fit the currently visible entities on load and whenever the visible set changes.

### Changes

- User experience
  - Map now zooms/centers to fit the visible entities instead of always showing the whole world.
  - Initial load snaps directly to the fitted view (no flash); subsequent changes to the visible entities animate to re-fit.
  
- Logical changes
  - Added `fitBounds(bounds, options)` to the `useMapZoom` hook: scales and centers the view on a bounding box (in map coordinates) with a configurable margin, clamped between the base (full-map) scale and the existing max zoom multiplier.
  - `EntityMap` computes the bounding box of the drawable entity centroids and calls `fitBounds` on load and on entity changes.

### Screenshots
- Before fully zoomed out
image

- Map now zooms
image


src/features/map/EntityMap.tsx
```diff
@@ -84,6 +91,46 @@ const EntityMap: React.FC<Props> = ({ entities, maxWidth = 2000, allowSidebar =
...
+    });
+
+    if (count === 0) return null;
+    return { minX, minY, maxX, maxY };
+  }, [drawableEntities, mapHeight]);
+
+  const hasInitialFitRef = useRef(false);
+  useEffect(() => {
+    if (!entityBounds) return;
+    // Instant on first load to avoid a flash from full-map to fitted; animate afterwards
```
conradarcturus, line 128: Thanks for paying attention to detail.

src/features/map/UseMapZoom.tsx
```diff
@@ -35,6 +35,21 @@ export type ZoomControlHandlers = {
...
+/** A bounding box expressed in map coordinates (same units as mapWidth/mapHeight). */
+export type MapBounds = {
+  minX: number;
+  minY: number;
+  maxX: number;
+  maxY: number;
+};
+
+type FitBoundsOptions = {
+  /** Fraction of the viewport to leave as margin around the bounds (0..1). Default 0.15. */
```
conradarcturus, line 47: very good idea

conradarcturus (approved):
This looks fantastic and the code is very solid. You are welcome to merge the code after the screenshots have been updated.

My only particular note is that the class files are becoming very large and there are no test classes evaluating the algorithm. I like to keep lambda functions <20 lines and files <100 lines but sometimes the refactor is more painful than useful so I leave it to you to decide.

btw I added the `update-baselines` tag so the screenshots are regenerated (rather than the test just failing)

PR #738, opened 2026-07-30
Shadcn: init Tailwind v4 and shadcn/ui on Base UI (base-mira)

hdprajwal:
Summary: Introduce Tailwind CSS v4 and shadcn/ui (Base UI, `base-mira` style) into the project. This is the foundation commit for the UI migration -- build wiring, theme tokens, and a single `Button` component. No other components are installed and no existing views are ported.

### Changes

- User experience
  - Intro page: objective inputs and the GO buttons now use Tailwind styling, with GO rendered by the new shadcn `Button`
  - Intro page: logo is centered again (Tailwind Preflight resets `img` to `display: block`, which defeated the container's `text-align: center`)
- Logical changes
  - `tailwindcss` + `@tailwindcss/vite` wired into `vite.config.ts`
  - `components.json`: `base-mira` on Base UI, `neutral` base color, CSS variables, lucide icons, generated components aliased to `@shared/ui`
  - `src/app/tailwind.css`: Tailwind import plus the mira token set (`:root` / `.dark`)
  - `cn()` helper at `@shared/lib/utils`
  - `@/*` alias added to `vite.config.ts`, `vitest.config.ts`, `tsconfig.app.json`, and `tsconfig.json` (the root config only holds references, so the shadcn CLI needs its own `paths` entry there)
  - `component_styles.css`: the global `button` rule is unlayered, so it beats Tailwind's `@layer utilities` and would override every shadcn component's own styling. Exempted `[data-slot]` elements, which shadcn sets and nothing in this codebase does.
- Data
  - None
- Refactors
  - None

Out of scope/Future work:

- Migrating existing views and components, and removing the legacy CSS (at which point the `:not([data-slot])` guard can be deleted)
- Tailwind Preflight resets are now live app-wide, so two regressions are visible and not yet addressed: paragraph margins are removed (Intro page paragraphs run together) and `h1`-`h6` render at body size ("Common Objectives" on Intro, "Core Pages" on About)
- Playwright baselines need regenerating in CI for the Preflight changes -- the screenshot job will be red until then
- The app keeps its own theming: no webfont added, `--font-sans` points at the existing font stack, and `colors.css` / `usePageBrightness` are untouched. `usePageBrightness` already adds a `dark` class for an explicit dark preference, so the shadcn `.dark` tokens work there; with `follow device` no class is set, so shadcn tokens stay on light values. Worth consolidating in a follow-up.

### Test Plan and Screenshots

How to test the changes in this PR: run `npm run dev` and open the [Intro page](http://localhost:5173/lang-nav/) -- the logo should be centered, and the objective inputs and GO buttons should show the new styling. Toggle light/dark via the settings menu. Then check the [Data page](http://localhost:5173/lang-nav/data) and [About page](http://localhost:5173/lang-nav/about) for unintended changes.

| Page/View with link | Description of Changes | Screenshot Before | Screenshot After |
| ------------------- | ---------------------- | ----------------- | ---------------- |
| [Intro](https://translation-commons.github.io/lang-nav/) | Logo centered; objective inputs and GO buttons on Tailwind / shadcn `Button` | | |
| [Data](https://translation-commons.github.io/lang-nav/data) | No intended change -- verified buttons, tabs, pagination, and filter chips render as before | | |
| [About](https://translation-commons.github.io/lang-nav/about) | No intended change other than the Preflight heading reset noted above | | |

# Checklist

## Summary

## Testing

- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test` -- 66 files, 419 tests passing
  - [ ] Tests added or updated for changed logic -- no behavior change to test; the new component is third-party generated code
- [x] `npm run dev` -- tried out the website directly, Intro / Data / About in both light and dark

### Visual changes

- [x] Screenshots to be attached to the PR (Intro page before/after)

### Internal changes

- [x] Build and config wiring only; the one non-obvious change (`:not([data-slot])`) is explained in a code comment


src/widgets/CommonObjectives.tsx
```diff
@@ -62,15 +64,15 @@ const Objective: React.FC<ObjectiveProps> = ({
...
+        className="inline-flex items-center"
         onSubmit={(e) => {
           e.preventDefault();
           window.location.href = `${urlPath}${getNewURL(params)}`;
         }}
       >
         {inputParam && (
           <input
-            style={{ padding: '0.25em 0.5em', marginLeft: '0.5em' }}
+            className="ml-2 h-7 rounded-md border border-border bg-transparent px-2 text-xs/relaxed outline-none transition-colors placeholder:text-muted-foreground focus-visible:border-ring focus-visible:ring-2 focus-visible:ring-ring/30"
```
conradarcturus: That's a crazy long class name, do we need all of it? I think it looks good with the regular styling and just a few touches:
```suggestion
            className="ml-2 p-2 rounded-md border"
```

src/widgets/CommonObjectives.tsx
```diff
@@ -87,10 +89,8 @@ const GoButton: React.FC<{ params: Partial<PageParams>; urlPath: LangNavPageName
   urlPath,
 }) => {
   return (
-    <a href={`${urlPath}${getNewURL(params)}`} style={{ marginLeft: '0.5em' }}>
-      <button style={{ padding: '0.25em 0.5em' }} type="submit">
-        GO
-      </button>
+    <a className="ml-2" href={`${urlPath}${getNewURL(params)}`}>
+      <Button type="submit">GO</Button>
```
conradarcturus, line 93: Do you know why this button is black on light-mode and white on dark-mode? It seems a bit stark.

conradarcturus (approved):
A lot of the buttons on LangNav are in a slightly broken state with this change. The most notable one is the tri-state page brightness selector. I think the additional button qualifier changed the CSS load order so some rules are overriding rules they didn't before. However, we do not need to resolve this immediately, we can follow up in time.

It may take awhile to fix each of the issues so I will actually accept this PR to unblock the work.

src/widgets/CommonObjectives.tsx
hdprajwal: the colors were inverted due to the shadcn default theme variables, replaced it with apps blue theme color

src/widgets/CommonObjectives.tsx
hdprajwal: I have removed some of the styling, this was to make sure we have the same styling across the shadcn buttons and out current buttons. 
Once we move over to using the shadcn buttons these classnames would be removed.

PR #743, opened 2026-08-02
Fix double scrollbar and detached sidebar on the Data page

### LINKED ISSUE #733: Fix double scrollbars // Figure out future of the data page panel layout
Under certain conditions there are extraneous scrollbars

quoted from feedback:

> I used LangNav today to search for the countries where languages are spoken. Here are a few observations: 
>
> I was running LangNav in a resizable window so I could have another app in a separate window alongside it. 
>
> With the resizable window there were 2 vertical scroll bars and one horizontal window. The inside scrollbar scrolled the list of languages. 
>
> I had to use the outside scroll bar to scroll the entire LangNav window down to get to the horizontal bar. So I have to scroll left to see the names and scroll up to see the language to select then scroll down to use the bar to slide right and scroll up to see the countries (and not seeing the names which were now off screen).  The UI should adjust so the outside scroll bar is not needed.

I think there is a broader existential question we need our designers to answer first. Right now the page body has 3 panels: Filter, Explore, & Details. Our current implementation of these causes the double scrollbars, but even further we need to figure out if we still want a triple panel situation or break up the page differently. I don't want to rush into an answer for this.

Shikhar16078:
Fixes #733

Summary: The Data page had a double scrollbar, the sidebar would visually detach from the navbar while scrolling, and the table's header row didn't stay pinned while scrolling through long lists. This PR makes the navbar sticky and sizes the sidebar and main content area to fit exactly between the navbar and footer, each scrolling internally on its own -- so there's a single clear scrollbar, the sidebar stays flush against the navbar, and the table header stays pinned.

### Test Plan and Screenshots

Open the [Data page](https://e89ee03c.lang-nav.pages.dev/data), scroll down on each view (Cards/Table/Hierarchy/Map/Reports), and confirm there's only one scrollbar, the navbar stays pinned, and the sidebar stays attached to it with no gap. In the Table view, scroll through a long list (e.g. Languages) and confirm the header row stays pinned below the navbar. Also check a narrow (~390px) viewport, since the navbar wraps to more rows there.

# Checklist

## Summary

- [x] Mention the issue

## Testing

- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test`
  - [x] Verified across all 5 Data page views, panel drag-to-resize, panel close/reopen, dark mode, and a 390px mobile viewport

### Visual changes

- [x] More views can be reproduced by just sharing the URL -- links to the relevant pages are included above.

### Internal changes

- [x] Logical changes


src/widgets/PageNavBar.tsx
```diff
@@ -12,9 +13,24 @@ import SettingsButton from './controls/SettingsButton';
 
 const PageNavBar: React.FC = () => {
   const { pageBrightness } = usePageParams().brightness;
+  const navRef = useRef<HTMLElement>(null);
```
conradarcturus: I'm not a huge fan of this ref to pass in the data -- but I don't have a better option, so why not.

src/features/layers/ZIndex.ts
```diff
@@ -5,6 +5,7 @@ const ZIndex = {
   MapSidePanel: 10,
 
   Sidepanel: 40,
+  NavBar: 41,
```
conradarcturus, line 8: Something about this change (this Zindex or something else) breaks the table view property of showing a sticky header row (notice the ZIndex for `TableStickyRow`) -- can you figure out how to maintain that functionality?

hdprajwal:
#root is already a flex column with three children. navbar, page, footer. the only reason it cannot work out the leftover height on its own is `min-height: 100vh`. because it is a min-height, the middle slot never gets a real height, so the data page cannot fill it. and this forces calc(100vh - ...), which needs the navbar and footer heights, which needs the two ResizeObservers and the two css vars.

you can change that one line and have the flexbox do the math

src/app/colors.css

```css
#root {
  display: flex;
  flex-direction: column;
  height: 100dvh;
}
```

src/app/App.tsx

```jsx
<div style={{ flex: 1, minWidth: 0, minHeight: 0, display: 'flex', flexDirection: 'column', overflow: 'auto' }}>
  <PageRoutes />
</div>
```

src/pages/DataPage.tsx

```jsx
<div style={{ display: 'flex', flex: 1, minHeight: 0 }}>
```

noticed you are using claude code on this one, so here is a prompt for it if that is easier than doing it by hand:

---

the double scrollbar fix on this branch measures the navbar and footer with ResizeObservers, publishes them as css vars, and sizes the data page off calc(100vh - those). #733 only asks for the outer window scrollbar to go away on the data page, so the table's horizontal scrollbar is reachable without scrolling the whole document.

look at #root in src/app/colors.css. it is already a flex column whose three children are the navbar, the page and the footer. but it is min-height, not height, so the middle slot never gets a definite height and nothing inside it can fill the leftover space. that is the only reason anything needs measuring.

let flexbox do the sizing instead, then delete everything the measuring approach needed. the panels and the data body should size off their container, not off the viewport. use dvh rather than vh so mobile browser chrome does not push the footer out of view.

leave the other pages alone. about, team and privacy should still scroll the window with the footer at the bottom of the document. only the intro and data page needs to fit the viewport exactly.

hdprajwal:
The updates look good to me, and the issues with the use of ResizeObservers is resolved.
@conradarcturus you can take a look at it once and approve the PR.

conradarcturus (approved):
Thanks for making the updates. We should use the tailwind classes -- I'll make the amendments for you.

PR #746, opened 2026-08-06
Shadcn: replace Pill with Badge and LoadingIcon with Spinner

hdprajwal:
Replaces two legacy components with their shadcn/ui equivalents.

## Components

| Removed | Replacement | Call sites |
| --- | --- | --- |
| `Pill.tsx` | `Badge` (`variant="secondary"`) | 9 |
| `LoadingIcon.tsx` | `Spinner` | 2 |

## Visual change

Badges render rounder and smaller: `Badge` is `rounded-full` with a fixed `h-5` and `text-[0.625rem]`, where `Pill` used `border-radius: 0.5em` and `font-size: 0.75em`. The background is unchanged — `Pill`'s `--color-button-secondary` is aliased onto `--secondary`, which is what `variant="secondary"` resolves to.

Screenshot baselines will need regenerating via the `update-baselines` label.

## Accessibility

`Spinner` ships `role="status"` and `aria-label="Loading"`, which `LoadingIcon` had no equivalent for. In `LoadingPage` the heading already announces the state, so the spinner is marked `aria-hidden` there to avoid a duplicate announcement. In `TableExport` the live region is kept, since nothing else conveys that an export is running.

`LoadingIcon` also injected a global `@keyframes spin` through an inline `<style>` tag on every render; `Spinner` uses Tailwind's built-in `animate-spin`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/features/table/TableExport.tsx
```diff
@@ -206,7 +206,11 @@ const ExportLabel: React.FC<{ exportType: ExportType; isExporting: boolean }> =
     case ExportType.Unchosen:
       return (
         <div style={{ display: 'flex' }}>
-          {isExporting ? <LoadingIcon /> : <ExternalLinkIcon className="button-inline-icon" />}{' '}
+          {isExporting ? (
+            <Spinner className="button-inline-icon" />
```
conradarcturus, line 210: I'm not sure you need the `button-inline-icon` class anymore.

Also the export button is generally having trouble these days, something with shadcn or tailwind changed the formatting to force it to take multiple lines.

conradarcturus (approved):
Fabulous. I have some concerns about the export button but that may have to be reworked later anyway.

PR #747, opened 2026-08-06
Shadcn: migrate the docs pages to Tailwind and shadcn

hdprajwal:
Stacked on #746.

Rebuilds the `/about` page family on Tailwind utilities and shadcn components. Every inline style in `pages/docs` and `widgets/docs` is gone, except `ColorSwatch`'s background, which stays dynamic because it paints whichever variable it documents.

## Pages covered

`/about`, `/team`, `/privacy-policy`, `/terms-of-use`, `/data-coverage`, `/data-sources`, `/code-style` — all now render zero inline styles, no legacy `.css` imports, and no raw HTML with a shadcn equivalent.

## Notable changes

- **`DocsCard` is now a shadcn `Card`.** The old component had no card chrome at all — just padding and a flex column — so both the navigation tiles and the prose sections gain a background and a ring. This is the largest visual change in the PR.
- **The component example table in `LinkComponents` is now a shadcn `Table`.**
- **`DocsCardGrid` had `gap: -0.5em 0`.** A negative gap is invalid CSS and was dropped by the browser, so the grid had no gap at all. It now has a real one.
- **`InternalLink` accepts a `className`**, so a linked `DocsCard` can be styled without falling back to inline styles. Additive — `style` still works for its other callers.
- **The Data Coverage table breaks out of the 800px prose column** from `2xl` up, so it centers rather than overflowing to the right. Below that breakpoint it stays in the column and scrolls horizontally.

## Second commit is separable

`Shadcn: use a grayscale foreground instead of navy` changes `--foreground` and its card/popover/secondary/accent/sidebar variants from `#04355c` to `#171717` (neutral-900), matching the `neutral` baseColor already in `components.json`. Dark mode already used `#f1f1f1` and is unchanged.

This one is **global, not scoped to the docs pages** — it affects every page and will move every screenshot baseline. It is a separate commit so it can be reviewed or dropped on its own.

## Deliberately out of scope

The components demoed on `/code-style` — `Hoverable`, `HoverableButton`, `ModalButton`, `PopupCard`, `Selector` — are untouched and still render as legacy inside the examples. Same for `FieldCoverageTable` on `/data-coverage`; only the page shell around it changed. Those belong to later layers.

`ColorStyles` still documents the legacy `--color-*` bridge variables, which are scheduled for deletion. Updating that list is a content change rather than a styling one, so it is left for a follow-up.

Screenshot baselines will need regenerating.


src/pages/docs/PrivacyPolicyPage.tsx
```diff
@@ -21,7 +21,7 @@ const PrivacyPolicyPage: React.FC = () => {
         </div>
         <div>
           This may include:
-          <ul style={{ margin: '0' }}>
+          <ul className="m-0">
```
conradarcturus: tailwind/shadcn borked the ul components, the `li`s are missing their pips(dots) now. Can you restore the list pips?

src/widgets/docs/DocsSection.tsx
```diff
@@ -2,11 +2,8 @@ import { ReactNode } from 'react';
 
 function DocsSection({ title, children }: { title: string; children: ReactNode }) {
   return (
-    <section
-      id={title.toLowerCase().replace(/\s+/g, '-')}
-      style={{ display: 'flex', flexDirection: 'column', gap: '0.5em', marginBottom: '1em' }}
-    >
-      <h2 style={{ margin: 0 }}>{title}</h2>
+    <section id={title.toLowerCase().replace(/\s+/g, '-')} className="mb-4 flex flex-col gap-2">
+      <h2 className="m-0">{title}</h2>
```
conradarcturus: Can you fix these h2 components to be bold and/or stand out in some way? They are flat with the text now, making it harder to read the docs.

src/widgets/docs/DocsCard.tsx
```diff
@@ -22,45 +24,46 @@ const DocsCard: React.FC<PropsWithChildren<Props>> = ({
...
-    opacity: isDisabled ? 0.72 : 1,
-  };
+  const isLinked = link != null && !isDisabled;
 
-  const body = (
-    <>
-      <div style={{ display: 'flex', alignItems: 'center', gap: '0.5em' }}>
-        <span style={{ fontWeight: 600 }}>{title}</span>
+  const card = (
+    <Card
```
conradarcturus, line 30: imho I don't like the border for these cards, but I am happy to follow your lead.

conradarcturus (changes requested):

conradarcturus (approved):

PR #752, opened 2026-08-11
Feat add more alt names

### LINKED ISSUE #734: Add more alternative names for languages from langtags.json
While not all SIL datasets are public domain, langtags.json is! When loading langnav we can fetch the data from <https://ldml.api.sil.org/langtags.json> and extract alternative spellings for languages. Read more about the dataset here: 
https://github.com/silnrsi/langtags/blob/master/doc/langtags.md -- note that it is organized by locales, so you may have to do some disambiguation or parsing. If there are any issues fetching from the URL we can include a copy of the json in the repository but we should check if there would be any issues keeping a copy.

You should load it in during the supplemental data step, and use the function `setLanguageNames` to add new names.

If there are new names, you'll see them appear when searching for languages and you should see them in the details view for a language.

kevinmdd:
Fixes #734

Summary: Adds SIL's langtags.json as a new supplemental data source to populate alternative language names (eg. "Castellano" / "Castilian" / "Español" for Spanish).

### Changes

- User experience
  - Additional alternative names now appear in language search and in the language details view for languages covered by the SIL langtags dataset (eg. Spanish, Mandarin Chinese, Yue Chinese, Kurmanji, Sorani, Standard Malay, Swahili, and others).
- Logical changes
  - langtags.json is organized per-locale (eg. es-Latn-ES, zh-Hans-CN), so multiple records can map to the same iso639_3 code. Names from all matching records are merged via setLanguageNames 
  - To avoid polluting macrolanguage entries with names that belong to a child language, loadLangTags skips adding names when the resolved language has LanguageScope.Macrolanguage scope. The specific language records in the same dataset independently supply the correct names to the correct node.
- Data
  - Added public/data/other_sources/langtags.json, a copy of SIL's langtags dataset (source: https://ldml.api.sil.org/langtags.json). Bundled locally as a fallback / for offline builds; see comment in loadLangTags.ts for details and re-fetch instructions.
- Refactors
  - None

Out of scope/Future work: Only iso639_3 name matching is handled, other langtags fields (eg. iana, localname, sldr) not used yet. 

### Test Plan and Screenshots

How to test the changes in this PR: Run the app and let supplemental data load, then search for "Spanish", click detailed view to show alternative names: Castellano, Castilian, Castillan, Espagnol, Español, Spanska. Next, search for "Mandarin Chinese" (cmn), which should show a full list of alternate names, and compare against "Chinese languages" (zho, the macrolanguage), which shouldn't show Mandarin-specific names like "Huayu" or "Kuo-Yu" and should only display generic naming. 

| Description of Changes | Screenshot Before | Screenshot After |
| ---------------------- | ----------------- | ---------------- |
|          Additional names shown for languages (e.g Spanish)        |    image | image |

# Checklist

Feel free to check off or just delete items in this section as you have completed them.

## Summary

- [X] Clear description of what and why
- [X] Scope kept focused; note follow-ups if any
- [X] Set yourself as assignee
- [X] Mention the issue (usually by writing "Fixes #ISSUE_NUMBER" or "Closes #ISSUE_NUMBER")

## Testing

- [X] `npm run lint`
- [X] `npm run build`
- [X] `npm run test`
  - [ ] Tests added or updated for changed logic
- [X] `npm run dev` -- tried out the website directly
  - [X] Include screenshots as noted below
  - [ ] Write comments on manual testing

### Visual changes

- [X] Add screenshots to the table template at the top of this file. You can include images inside the table
  - [X] Drag and drop images in the GitHub PR comment box to upload screenshots

### Data changes

- [X] TSV, SVG, etc. edits in `public/data/`
- [X] Corresponding readmes updated in `public/data/`
- [X] Load/connect/compute updates in `src/features/data/` including how we aggregate data or compute derived values

### Internal changes

- [X] Logical changes
- [ ] Refactors, moving files around
- [ ] If you notice any changes that require explanations, make sure to include the explanations in the code as well.

## Docs

- [X] Code is self-documenting, or if not, comments are added where needed.
- [ ] Updated markdown readme files documenting how the code behaves or how to develop in case there are any relevant changes to make.


hdprajwal:
Nice work on the change, the code follows exsting patterns,
The newly added JSON file is around 4.6 MB and we only use two fields from it. The loader reads iso639_3 and names which means every visitor to the data page downloads and parses it. 
I would suggest writing a simple script that can convert this json data into tsv data of only the fields that we use.
you can use ';' as the separator for individual names (verified that ; does not appear in the data anywhere).

hdprajwal:
Some languages have the same names that appear twice with different capitalization.

image

We want to convert all the names to lowecase when comparing to get the unique names.
We would not need to lowercase the output itself. Names like eBhele and shiChopi have meaningful capital letters.

src/features/data/load/supplemental/loadLangTags.ts
```diff
@@ -0,0 +1,39 @@
...
+          return;
+        }
+
+        const lang = getLanguage(language.iso639_3);
+        if (lang === undefined) {
+          return;
+        }
+
+        if (lang.scope === LanguageScope.Macrolanguage) {
+          return;
```
conradarcturus: 2 things: have shorter bailout statements

and for the macrolanguage name problem, use `iso639_3extra` to figure out if the row is for a specific language. Thanks for noticing that!

```suggestion
        const lang = getLanguage(langtag.iso639_3extra ?? langtag.iso639_3);
        if (!lang) return;
```

src/features/data/load/supplemental/loadLangTags.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function loadLangTags(
+  getLanguage: (id: string) => LanguageData | undefined,
+): Promise<void> {
+  return fetch('data/other_sources/langtags.json')
+    .then((res) => res.json())
+    .then((languages: SILLanguage[]) => {
+      languages.forEach((language) => {
+        if (!language.iso639_3) {
+          return;
+        }
```
conradarcturus: I prefer compact bailouts
```suggestion
        if (!language.iso639_3) return;
```

src/features/data/load/supplemental/loadLangTags.ts
```diff
@@ -0,0 +1,39 @@
+// Source: https://ldml.api.sil.org/langtags.json (SIL langtags dataset, public domain)
+// Docs: https://github.com/silnrsi/langtags/blob/master/doc/langtags.md
+// A local copy is bundled at data/other_sources/langtags.json as a fallback / for
+// offline builds. If refreshing this file, re-fetch from the URL above.
+
+import { LanguageData, LanguageScope } from '@entities/language/LanguageTypes';
+import { setLanguageNames } from '@entities/language/setLanguageNames';
+
+export interface SILLanguage {
+  iso639_3?: string;
```
conradarcturus: Technically it is a LangTag (~= Locale) not a Language.

Also add the `iso639_3extra`

```suggestion
export interface SILLangTag {
  iso639_3?: string;
  iso639_3extra?: string;  
```

conradarcturus (changes requested):
Thanks for adding this! simple and sweet. I can see the new names.

@hdprajwal Thanks for helping to review :) For the filesize I have mixed feelings. I like the idea of keeping the input data similar to the way it is exported (easier to update, easier to extend) but I sympathize with the data size concern.

@kevinmdd I leave it to you how much you want to do that. It shouldn't be too bad to make a script to convert it to a TSV of iso639-3 -> name list. Make sure to include that file in the repo as well so we can re-run it after the langtag repo updates. If this is too much of a headache we can make this a followup task.

I'm hoping the backend team is able to finish their exploration so we don't have to load in all of the source data files. Rather if we have the cached & indexed versions of LanguageData[] etc so we can avoid concerns about input dataset size. There is other rich data in langtags.json that we may also want to add to LangNav (at a later date), like likely locale codes.

kevinmdd:
Thank you @conradarcturus @hdprajwal for the feedback, it should be ready to merge to master! I had originally thought of changing it from JSON to TSV but I wasn't entirely sure if it was something we had needed to keep as a JSON. Glad I got the confirmation.

public/data/other_sources/langtags.tsv
```diff
@@ -0,0 +1,9301 @@
...
+akj		Jero;Jeru;Yerawa
+akk		
+akl		Akeanon;Aklan;Aklano;Aklanon-Bisayan;Inakeanon;Panay;Qinakłanón
+akm		Ba;Bo
+ako		Akoerio;Akuliyo;Akuri;Akurijo;Akuriyo;Oyaricoulet;Triometesem;Triometesen;Wama;Wayaricuri
+akp		Akpafu-Lolobi;Lolobi-Akpafu;Mawu;Siwusi
+akq		
+akr		Araki;Ravoraki
+aks		Akasele;Akaselecɛ;Bikaselebíyáá;Cemba;Chamba;Kamba;Kasele;Kassele;Tchamba;àkasìlìmí
+akt
```
conradarcturus: Let's have the script remove empty rows

public/data/other_sources/langtags.tsv
```diff
@@ -0,0 +1,9301 @@
...
+ale		Anangax;Unangam tunnu;Unangan;Unangany;Unangax;Unanghan
+alf		Alege;Alegi;Eligi;Ugbe;Uge
+alh		Alaua;Allaura;Allawa;Allowa;Allowiri;Allua;Alowa;Galawa;Galleewo;Kallana;Kallaua;Leealowa;Warliburru
+ali		Okim
+alj		
+alk		Halak;Harak;Harlak;Hrlak
+all		Aalan;Alan;Alanmar;Alar;Allan;Chatans
+alm		Aje;Varavara
+aln		Geg;Gegnisht;Gegnisht-Shqyp;Gheg;Gheg Albanian;Guegue;Shopni;Shqip;Shqyp
+aln		Geg;Gegnisht;Gegnisht-Shqyp;Gheg;Gheg Albanian;Guegue;Shopni;Shqip;Shqyp
```
conradarcturus: also don't have it add duplicated rows.

src/entities/language/setLanguageNames.ts
```diff
@@ -14,12 +14,14 @@ export function setLanguageNames(lang: LanguageData, additionalNames: string[] =
   // Remove parentheticals if its a standard for an alternative language name
   // eg. ["Mandarin", "Putonghua (Mandarin)"] -> ["Mandarin", "Putonghua"]
   // but keep if its is an elaboration eg. ["Cantonese", "Cantonese (inc. Teochew, Taiwanese)"]
-  const namesRemovingParentheticalDuplicates = unique(
+  const namesRemovingParentheticalDuplicates = uniqueBy(
```
conradarcturus, line 17: perfect

scripts/langtags-conversion.mjs
```diff
@@ -0,0 +1,24 @@
+import fs from 'node:fs';
```
conradarcturus, line 13: Add how to run instructions

conradarcturus (approved):
Some minor changes that you can resolve now or in a followup.

PR #753, opened 2026-08-12
Responsive Navbar

conradarcturus:
We recently improved the page layout, mitigating multiple scrollbars. However one problem I saw was that the navbar was not responsive. If the page was narrow the navbar could occupy much of the screen. Thereby, this change enables the navbar to accomodate various widths (see screenshots below).

Note: I did not use shadcn's library navbar components because it was just so different than our current navbar.

No AI was used to author the original commit in this PR because it often hallucinates styles

### Changes

- User experience
  - NavBar changes format on tailwind media intervals, like the existence/length of the langnav title or if the search is opened via a button or always present
  - Removed "Intro" button, it was redundant with the main link
  - Added Tools menu button to quickly access recommended pages that are embedded within LangNav
- Logical changes
  - n/a
- Data
  - n/a
- Refactors
  - Lots of style declarations moved to use tailwind
  - Some improvements to Popupcard with the lessons learned from NewHoverable

Out of scope/Future work: ...

### Test Plan and Screenshots

How to test the changes in this PR: Used devtools to try various widths

| Page/View with link  | Screenshot Before | Screenshot After |
| ------------------- | ----------------- | ---------------- |
|Worst case, <350px|Screenshot 2026-08-11 at 20 27 33|Screenshot 2026-08-11 at 20 27 13
|Small case, 464px|Screenshot 2026-08-11 at 20 32 45|Screenshot 2026-08-11 at 20 28 25|
|Desktop case, 1200px|Screenshot 2026-08-11 at 20 35 56|Screenshot 2026-08-11 at 20 34 46

Searchbar popup
Screenshot 2026-08-11 at 20 37 30

Tools popup (with plural forms opened)
Screenshot 2026-08-11 at 20 37 51


hdprajwal:
This is a much needed change to make the site responsive, I have a couple of suggestions and feedback.

1. Since in the mobile view we can double the navbar height and show both the full search bar and the feedback, settings buttons.
2. PopupCard previously closed on Escape and on the X button. Components like the Feedback and setting popupcards now can not be dismissed using escape keys.
3. SearchBar is mounted twice at every screen width, once inside the popup and once in the lg:flex div, so the entity list gets filtered and sorted twice on every navbar render instead of once

